PR #23, opened 2023-01-11
Wrong sub command name for stack

brunowego:
Currently in the README file, the command for create a sandbox is wrong.


flemzord:
Thanks @brunowego !

PR #55, opened 2023-02-13
fix: add proper version for gateway for local dev and bump deps

paul-nicolas:
Now that the gateway is merged and released, we have a proper version to put for local dev and also for the builder config to build caddy with the plugins.

Because of a breaking change in opentelemetry (removal of packages) caddy is not buildable anymore with the version v2.6.2. We had to upgrade to v2.6.3.
Because of the go.work.sum, the open telemetry breaking change was applied to the monorepo and I had to bump all deps depending on it in every components

<a href="https://gitpod.io/#https://github.com/formancehq/stack/pull/55"></a>


flemzord (approved):

flemzord (approved):

paul-nicolas:
PR tested in local with local versions, ready to merge

PR #86, opened 2023-02-22
fix: tests with postgres server

gfyrag:
The recently added tests inside cmd/ was not starting a dedicated pg server instance, breaking a simple "go test" which was working on all other tests.
Also, changes really check errors of the commands.
And changes fix viper missing flags (was working only because tests inside root_test.go was configuring global env var).

### DIFF
--- a/components/ledger/cmd/container_test.go
+++ b/components/ledger/cmd/container_test.go
@@ -12,7 +12,7 @@ import (
 
 	"github.com/ThreeDotsLabs/watermill/pubsub/gochannel"
 	"github.com/formancehq/stack/libs/go-libs/otlp/otlptraces"
-	"github.com/numary/ledger/internal/pgtesting"
+	"github.com/formancehq/stack/libs/go-libs/pgtesting"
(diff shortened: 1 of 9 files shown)


antoinegelloz (approved):

components/ledger/cmd/container_test.go
```diff
@@ -12,7 +12,7 @@ import (
 
 	"github.com/ThreeDotsLabs/watermill/pubsub/gochannel"
 	"github.com/formancehq/stack/libs/go-libs/otlp/otlptraces"
-	"github.com/numary/ledger/internal/pgtesting"
+	"github.com/formancehq/stack/libs/go-libs/pgtesting"
```
paul-nicolas, line 15: Should we do a pass for all ledger and generally all components imports to have every imports pointing to the stack monorepo ? (Not in this PR of course)

gfyrag: Yes, i think we should.
Do you have another idea?

paul-nicolas (approved):

PR #88, opened 2023-02-23
test: add more tests

gfyrag:
Changes add : 
* some checks on accounts and assets availabilty on search service after a transaction creation.
* some checks on accounts metadata availability on search service after a configuration of an account metadata.
* fix the sdk for listing payments (the openapi was missing a payment status enum case)


events/README.md
```diff
@@ -0,0 +1,9 @@
+# Events
+
+This repository centralize events schemas across Formance Stack.
+For each stack version, a repository "vX" contains all related events.
+
+Each "vX" folder contains a "base" folder containing the base event format, common to all services.
+This base event contains a "type" property, and an "app" property, denoting the format of the "payload" property.
+
+For example, an event with "type" == "SAVED_PAYMENT" and app == "payments", must have a payload matching schema in file payments/SAVED_PAYMENT.yaml.
```
antoinegelloz, line 9: Small mistakes :)
```suggestion
This repository centralizes event schemas across the Formance Stack. For each stack version, a repository named "vX" contains all related events.

Each "vX" folder contains a "base" folder that contains the base event format, which is common to all services. This base event includes a "type" property and an "app" property, which denote the format of the "payload" property.

For example, an event with "type" == "SAVED_PAYMENT" and "app" == "payments" must have a payload matching schema in the file "payments/SAVED_PAYMENT.yaml".
```

events/events.go
```diff
@@ -0,0 +1,58 @@
+package events
```
paul-nicolas, line 1: Maybe we should put this in the /libs directory ? (just my personal opinion on this point)

paul-nicolas (approved):

PR #121, opened 2023-03-08
feat: replace goose by handcradfted library

gfyrag:


components/orchestration/cmd/serve.go
```diff
@@ -42,7 +42,7 @@ func newServeCommand() *cobra.Command {
 				fx.Invoke(func(lifecycle fx.Lifecycle, db *bun.DB) {
 					lifecycle.Append(fx.Hook{
 						OnStart: func(ctx context.Context) error {
-							return storage.Migrate(db, viper.GetBool(service.DebugFlag))
+							return storage.Migrate(db)
```
paul-nicolas, line 45: Why removing the debug flag ?

paul-nicolas: I said nothing, it was for goose, which is removed now

gfyrag: Goose support this, i have not reimplemented it.
Do you think I should?

paul-nicolas: It depends if we rely on it to debug migrations at first, but in any case I would say yes we should reimplement it (maybe later if needed if you're block for other things)

paul-nicolas (approved):

antoinegelloz (approved):

paul-nicolas (approved):

PR #126, opened 2023-03-09
feat: Handle CONNECTOR_RESET event on search ingestion"

gfyrag:


paul-nicolas (approved):
Discussed and explained by call, lgtm !

PR #128, opened 2023-03-10
feat: introduce logs insertion by batch

paul-nicolas:
BEWARE: batch size must stay at 1 until we have implemented the account based lock and CQRS for transactions, accounts and balances.

### DIFF
--- a/components/ledger/go.mod
+++ b/components/ledger/go.mod
@@ -33,6 +33,7 @@ require (
 	go.opentelemetry.io/otel/sdk v1.13.0
 	go.opentelemetry.io/otel/trace v1.13.0
 	go.uber.org/fx v1.19.2
+	golang.org/x/sync v0.1.0
 	gopkg.in/segmentio/analytics-go.v3 v3.1.0
 )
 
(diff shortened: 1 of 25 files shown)


components/ledger/pkg/storage/sqlstorage/worker/worker.go
```diff
@@ -0,0 +1,111 @@
...
+		workerJob: workerJob,
+
+		models:     nil,
+		modelsChan: make(chan []Model[MODEL], 1024),
+	}
+}
+
+// Run should be called in a goroutine
+func (w *Worker[MODEL]) Run(ctx context.Context) {
+	w.ctx = ctx
```
paul-nicolas, line 40: Let's not use it in the struct

components/ledger/go.mod
```diff
@@ -35,6 +35,7 @@ require (
 	go.opentelemetry.io/otel/sdk v1.13.0
```
Misty0411, line 35: [
```suggestion
 	go.opentelemetry.io/otel/sdk # v1.13.0

flemzord: @Misty0411 Why this modification ?

gfyrag (approved):

components/ledger/go.mod
Misty0411: Its not me

PR #129, opened 2023-03-13
feat: move ledger lock inside core ledger

gfyrag:


components/ledger/pkg/ledger/execute_script.go
```diff
@@ -21,6 +21,13 @@ import (
 )
 
 func (l *Ledger) ExecuteScript(ctx context.Context, preview bool, script core.ScriptData) (core.ExpandedTransaction, error) {
+
+	unlock, err := l.locker.Lock(ctx, l.store.Name())
```
paul-nicolas, line 25: Should we check that the locker is nil ? (Even if it's in the default options)

antoinegelloz (approved):

components/ledger/pkg/ledger/resolver.go
```diff
@@ -104,10 +107,13 @@ func ProvideResolverOption(provider interface{}) fx.Option {
...
 		fx.Provide(
-			fx.Annotate(func(storageFactory storage.Driver[Store], ledgerOptions []LedgerOption, options ...ResolverOption) *Resolver {
-				return NewResolver(storageFactory, ledgerOptions, cacheBytesCapacity, cacheMaxNumKeys, options...)
-			}, fx.ParamTags("", ResolverLedgerOptionsKey, ResolverOptionsKey)),
+			fx.Annotate(func(storageFactory storage.Driver[Store], ledgerOptions []LedgerOption, locker Locker, options ...ResolverOption) *Resolver {
+				return NewResolver(storageFactory, ledgerOptions, cacheBytesCapacity, cacheMaxNumKeys, locker, options...)
+			}, fx.ParamTags("", ResolverLedgerOptionsKey, "", ResolverOptionsKey)),
 		),
+		fx.Provide(func() Locker {
+			return NewInMemoryLocker()
```
paul-nicolas, line 115: Ok, so no redis anymore since we are in monowriter ? Should we keep it in case we need it one day ?

paul-nicolas (approved):

components/ledger/pkg/ledger/resolver.go
gfyrag: I don't think is is necessary.
If we use Redis later, the implementation for account based lock will be more complicated.

components/ledger/pkg/ledger/execute_script.go
gfyrag: I don't think so, il a kind of "defensive programming" :)
But i can if you prefer.

paul-nicolas: No i'm good with keeping that like that if we are sure that it will never be nil !

PR #145, opened 2023-03-14
feat: move bulk insert to copy with bun

paul-nicolas:
Bulk insert were done thanks to bun and mulitple
values for the insert postgres method. In order to improve the performances, let's move to copy for
these bulk inserts.

Also, using copy wiuth bun is not working if the pgx driver is used when creating the bun db.


components/ledger/pkg/storage/sqlstorage/ledger/transactions.go
```diff
@@ -399,13 +399,15 @@ func (s *Store) insertTransactions(ctx context.Context, txs ...core.ExpandedTran
 		Model(&ps).
 		Exec(ctx)
 	if err != nil {
+		fmt.Println("TOTOTOTOTO", err)
```
gfyrag, line 402: :)

gfyrag (approved):

PR #153, opened 2023-03-16
feat(operator): helm deployment initialization

louhde:


components/operator/config/rbac/role.yaml
```diff
@@ -1,9 +1,10 @@
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
```
gfyrag, line 3: Are you sure it works?
The operator use some cluster scoped resources.

louhde: Didn't sure .. I take a look why Kubebuilder changed it

components/operator/config/rbac/role.yaml
```diff
@@ -132,3 +132,15 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
```
gfyrag, line 135: Can be removed (not a blocker).

louhde: I prefer to keep it I think I will need it in the future

components/operator/Makefile
```diff
@@ -102,6 +102,37 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
 	$(KUSTOMIZE) build config/default | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
 
+.PHONY: helm-update
+helm-update: manifests generate
```
flemzord, line 106: Maybe need to exclude Kutomiszation.yaml ?

components/operator/internal/controllers/stack/stack_reconciler.go
```diff
@@ -42,6 +42,7 @@ const (
 // +kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=cert-manager.io,resources=certificates,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=traefik.containo.us,resources=middlewares,verbs=get;list;watch;create;update;patch;delete
```
flemzord: Remove this line because i don't need this.

louhde: Done

components/operator/Makefile
louhde: I did it in .helmignore and .gitignore

PR #155, opened 2023-03-17
feat: switch bus messages to protobuf

paul-nicolas:
This will allow us to have some versionning of messages inside the buses (kafka, nats etc...) in order to not have any surpirse in the future.


paul-nicolas:
c.f.: https://www.notion.so/formance/JSON-Protobuf-in-buses-cc811127814f4fbe978b44f57df1a632?pvs=4

PR #156, opened 2023-03-17
feat: prepare multi account lock interface

gfyrag:


components/ledger/pkg/api/controllers/transaction_controller.go
```diff
@@ -222,15 +224,17 @@ func (ctl *TransactionController) PostTransaction(w http.ResponseWriter, r *http
...
+	res, logHandler, err := l.CreateTransaction(r.Context(), preview, script)
 	if err != nil {
 		apierrors.ResponseError(w, r, err)
 		return
 	}
 
-	if err := logs.Wait(r.Context()); err != nil {
-		apierrors.ResponseError(w, r, err)
-		return
+	if !preview {
```
paul-nicolas, line 233: Normally it's automatically handled by the fact that the log handler that is returned is empty !

paul-nicolas (approved):

components/ledger/pkg/api/controllers/transaction_controller.go
paul-nicolas: Alright

PR #122, opened 2023-03-08
feat: sqlstorage rework

paul-nicolas:
## Description

This PR does multiple things:
- Separate system and ledger schema
- Move every sql query to uptrace/bun instead of sqlbuilder
- Fix all tests
- Remove all sql transactions and the middleware associated (beware that there is no lock right now while waiting to re-implement one)

PR #116, opened 2023-03-06
fix: local testing

gfyrag:

PR #160, opened 2023-03-20
feat: refactor commit process

gfyrag:
- git subrepo push components/ledger
- feat(ledger): replace gin by chi and remove authorization options (#92)
- feat: Remove NUMARY_ prefix on ledger and update operator (#94)
- feat: move openapi files to root (#95)
- feat: remove ledger config file (#96)
- feat: remove deprecated flags and add --bind flag on ledger (replace --server.http.bind_address)" (#97)
- feat: rename references from numary to ledger (#98)
- feat: remove deprecated json fields and query params (#100)
- feat/num 1546 ledger replace server start by serve (#101)
- feat: remove /script endpoint (#102)
- feat: remove mapping and batch (#104)
- chore: remove 'ui' and 'stickers' commands on ledger (#105)
- chore: remove http basic authentication and deprecated telemetry flags (#106)
- chore: remove 'total' property from cursors (#107)
- feat: remove wrapper on GET /_info (#108)
- feat: remove sqlite usage on auth server (#110)
- feat: remove sqlite from ledger (#109)
- fix: local testing (#116)
- feat: use pgtesting from go libs (#112)
- fix: docker-compose entrypoints (#118)
- feat(orchestration): add --worker flag to embed worker on main service (#120)
- feat: replace goose by handcradfted library (#121)
- feat: sqlstorage rework (#122)
- feat: move ledger lock inside core ledger (#129)
- feat: introduce logs insertion by batch (#128)
- operator/tests: fix missing NUMARY prefix
- fix: if we have an error, still release all models (#141)
- feat: move bulk insert to copy with bun (#145)
- feat: combine logs before insertion (#142)
- feat: add openapi spec validation on integration tests (#133)
- feat: remove unnecessary storage abstractions (#147)
- feat: add basic cqrs implementation
- feat: enforce CQRS
- feat: prepare multi account lock interface
- refactor: commit process


components/ledger/pkg/ledger/cache/cache.go
```diff
@@ -0,0 +1,76 @@
...
+		account, err := c.store.ComputeAccount(ctx, address)
+		if err != nil {
+			return nil, err
+		}
+
+		if err := c.cache.Set(c.accountKey(account.Address), account); err != nil {
+			panic(err)
+		}
+
+		return account, nil
```
paul-nicolas, line 33: should we copy also here before returning the cache ?

components/ledger/pkg/ledger/lock/lock.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	d.globalLock.Lock()
+	lock, ok = d.locks[ledger] // Double check, the lock can have been acquired by another go routing between RUnlock and Lock
+	if !ok {
+		lock = &sync.Mutex{}
+		d.locks[ledger] = lock
+	}
+	d.globalLock.Unlock()
+ret:
```
paul-nicolas, line 43: I'm not a big fan of goto, is it better to put the previous lock instead of the goto ?

components/ledger/pkg/ledger/query/worker.go
```diff
@@ -0,0 +1,235 @@
...
+	} else {
+		v := uint64(0)
+		nextTxID = &v
+	}
+
+	for _, log := range logs {
+		var err error
+		switch log.Type {
+		case core.NewTransactionLogType:
+			payload := log.Data.(core.NewTransactionLogPayload)
```
paul-nicolas, line 135: Should we separate in functions the switch ? Not a blocker for me if you prefer like this, but maybe it will be clearer to separate ? (example: processTransactionLog, processSetMetadata etc...)

components/ledger/pkg/ledger/resolver.go
```diff
@@ -88,39 +61,26 @@ func (r *Resolver) GetLedger(ctx context.Context, name string) (*Ledger, error)
...
-	return NewLedger(store, r.monitor, r.cache, append(r.ledgerOptions, WithLocker(r.locker))...)
-}
-
-func (r *Resolver) Close() {
-	r.cache.Close()
-}
-
-const ResolverOptionsKey = `group:"_ledgerResolverOptions"`
-const ResolverLedgerOptionsKey = `name:"_ledgerResolverLedgerOptions"`
+ret:
```
paul-nicolas, line 64: same comment for the goto

components/ledger/pkg/bus/aggregate.go
```diff
@@ -0,0 +1,23 @@
+package bus
+
+import (
+	"github.com/formancehq/ledger/pkg/core"
+)
+
+func aggregatePostCommitVolumes(txs ...core.ExpandedTransaction) core.AccountsAssetsVolumes {
```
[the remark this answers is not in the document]
gfyrag: It could be but, today, it is used only in this package.

components/ledger/pkg/ledger/cache/cache.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return err
+	}
+	account.Metadata = account.Metadata.Merge(m)
+	_ = c.cache.Set(c.accountKey(address), account)
+	return nil
+}
+
+func (c *Cache) accountKey(address string) string {
+	return c.store.Name() + "-" + address
```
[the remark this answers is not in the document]
gfyrag: Good catch, it was global on an earlier version.

components/ledger/pkg/ledger/cache/cache.go
gfyrag: Good catch!

components/ledger/pkg/ledger/lock/lock.go
gfyrag: The previous lock is a read lock.
But i an modify this to remove the goto if you don't like it.

components/ledger/pkg/ledger/resolver.go
gfyrag: Ok, i'll change this

PR #161, opened 2023-03-21
feat: add internal channel for views computing

paul-nicolas:
This commit adds an internal channel for views computing and also changed the type when logs are computed to have a pointer. This will allow us to retrieve the log ID when it will be inserted in order to store it in the query worker.


components/ledger/pkg/ledger/query/worker.go
```diff
@@ -33,14 +53,18 @@ func (w *Worker) Run(ctx context.Context) error {
...
 			return nil
-		case <-time.After(w.Interval):
-			if err := w.run(ctx); err != nil {
+		case log := <-w.logChan:
+			if err := w.processLog(ctx, log.store, log.log); err != nil {
 				if err == context.Canceled {
 					logging.FromContext(ctx).Debugf("CQRS worker canceled")
 					return err
 				}
 				logging.FromContext(ctx).Errorf("CQRS worker error: %s", err)
```
gfyrag, line 62: In case of error, i think we should not update the next log id. Either we crash the app to allow the polling part to make its job, either the process need a way to retry later.

components/ledger/pkg/ledger/ledger.go
```diff
@@ -41,9 +50,14 @@ func (l *Ledger) GetLedgerStore() storage.LedgerStore {
 }
 
 func (l *Ledger) CreateTransaction(ctx context.Context, dryRun bool, script core.RunScript) (*core.ExpandedTransaction, error) {
-	return l.runner.Execute(ctx, script, dryRun, func(expandedTx core.ExpandedTransaction, accountMetadata map[string]core.Metadata) core.Log {
+	tx, log, err := l.runner.Execute(ctx, script, dryRun, func(expandedTx core.ExpandedTransaction, accountMetadata map[string]core.Metadata) core.Log {
 		return core.NewTransactionLog(expandedTx.Transaction, accountMetadata)
 	})
+	if err == nil {
+		err = l.queryWorker.QueueLog(ctx, log, l.store)
```
gfyrag, line 57: I think QueueLog() should not return any error.
In this case, even if we return an error, the log will be written on disk.

components/ledger/pkg/ledger/query/worker.go
paul-nicolas: Instead of crashing the app, what do you think about crashing the worker goroutine and relaunching it ?

components/ledger/pkg/ledger/ledger.go
paul-nicolas: Since we have a way to compute it again with the cache, i'm not against doing nothing on this case. Do you think just removing the error will be fine on your side ?

gfyrag (approved):

PR #174, opened 2023-03-23
refactor: commit write buffer

gfyrag:


components/ledger/pkg/storage/sqlstorage/worker/worker.go
```diff
@@ -2,110 +2,137 @@ package worker
...
+		case <-w.releasedJob:
+			if len(w.pending) > 0 {
+				for {
+					select {
+					case <-ctx.Done():
+						return
+					case ch := <-w.stopChan:
+						close(ch)
+						return
+					case w.jobs <- w.pending:
```
paul-nicolas, line 87: Should we not make a copy of the pending array ? It will be erase with the next line am I right ?

components/ledger/pkg/storage/sqlstorage/worker/worker.go
```diff
@@ -2,110 +2,137 @@ package worker
...
+					}
+				}
+			}
+			select {
+			case <-ctx.Done():
+				return
+			case ch := <-w.stopChan:
+				close(ch)
+				return
+			case mh := <-w.writeChannel:
```
paul-nicolas, line 99: Why are you sending only one models just after ? Why not start the loop again ?

components/ledger/pkg/storage/sqlstorage/worker/worker.go
paul-nicolas: Or maybe not, discussed IRL !

paul-nicolas (approved):

components/ledger/pkg/storage/sqlstorage/worker/worker.go
paul-nicolas: Discussed IRL, maybe just a comment on it will be clearer in the future !

paul-nicolas (approved):

PR #187, opened 2023-03-29
feat: optimize queries package

paul-nicolas:
This PR does multiple things:
- Switch the query worker to have one per ledger instantiated instead of a global one
- Process multiple logs at the same time 
- Adds a transaction when inserting everything


components/ledger/pkg/ledger/monitor/monitor.go
```diff
@@ -15,10 +16,13 @@ type Monitor interface {
 type noOpMonitor struct{}
 
 func (n noOpMonitor) CommittedTransactions(ctx context.Context, s string, res ...core.ExpandedTransaction) {
+	fmt.Println("FAKE MONITOR COMMITED TRANSACTIONS")
```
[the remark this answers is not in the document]
paul-nicolas: My bad, i'm going to remove it !

components/ledger/pkg/ledger/query/worker.go
```diff
@@ -184,125 +291,204 @@ func (w *Worker) initLedger(ctx context.Context, ledger string) error {
...
-		if err := store.InsertTransactions(ctx, expandedTx); err != nil {
-			return errors.Wrap(err, "inserting transactions")
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	if w.monitor != nil {
+		for _, monitor := range logsData.monitors {
+			monitor(ctx, w.monitor)
```
gfyrag, line 337: maybe run in a routine?

gfyrag (approved):

components/ledger/pkg/ledger/query/worker.go
paul-nicolas: I'm not a big fan of that since we can have the events not in order ! After discussing IRL, we'll let that like this for now !

PR #193, opened 2023-03-30
feat: implement multi account based lock

gfyrag:


components/ledger/pkg/ledger/ledger.go
```diff
@@ -174,8 +175,9 @@ func (l *Ledger) SaveMeta(ctx context.Context, targetType string, targetID inter
 
 	at := core.Now()
 	var (
-		err error
-		log core.Log
+		err     error
+		log     core.Log
+		release cache.Release = func() {}
```
paul-nicolas, line 180: Empty func for cache.Release ?

paul-nicolas (approved):

components/ledger/pkg/ledger/ledger.go
gfyrag: Yes. For transactions, we don't lock anything, so the release function is juste empty, this just avoid a "if"...

PR #195, opened 2023-03-30
chore: Orchestration integration tests

antoinegelloz:


gfyrag (approved):

components/orchestration/internal/workflow/instance.go
```diff
@@ -15,7 +15,7 @@ type Instance struct {
 	UpdatedAt     time.Time  `json:"updatedAt"`
 	Terminated    bool       `json:"terminated"`
 	TerminatedAt  *time.Time `json:"terminatedAt,omitempty"`
-	Statuses      []Stage    `json:"status,omitempty" bun:"rel:has-many,join:id=instance_id"`
+	Statuses      []Stage    `json:"statuses,omitempty" bun:"rel:has-many,join:id=instance_id"`
```
gfyrag, line 18: Can you keep "statuses" until feat/v2 is merged?
Maybe just set a TODO.
Because @reslene is actively working on orchestration, and it will lead to a breaking change we will forgot.

components/orchestration/openapi.yaml
antoinegelloz, line 1: @gfyrag There was a required field `statuses` here that was never returned by the API

tests/integration/suite/orchestration-workflows-execute-with-error.go
```diff
@@ -0,0 +1,99 @@
...
+					RequestBody(map[string]string{}).
+					Execute()
+				Expect(err).ToNot(HaveOccurred())
+			})
+			Then("waiting for first stage retried at least once", func() {
+				var getWorkflowInstanceHistoryStageResponse *formance.GetWorkflowInstanceHistoryStageResponse
+				BeforeEach(func() {
+					Eventually(func(g Gomega) int32 {
+
+						//TODO: error EOF response
```
antoinegelloz, line 62: @gfyrag test to fix here

tests/integration/suite/orchestration-workflows-execute.go
```diff
@@ -0,0 +1,219 @@
...
+								},
+							}))
+							Expect(getWorkflowInstanceHistoryStageResponse.Data[0].StartedAt).NotTo(BeZero())
+							Expect(getWorkflowInstanceHistoryStageResponse.Data[0].TerminatedAt).NotTo(BeZero())
+							Expect(getWorkflowInstanceHistoryStageResponse.Data[0].LastFailure).To(BeNil())
+							Expect(getWorkflowInstanceHistoryStageResponse.Data[0].StartedAt).NotTo(BeZero())
+							Expect(getWorkflowInstanceHistoryStageResponse.Data[0].Attempt).To(Equal(int32(1)))
+							Expect(getWorkflowInstanceHistoryStageResponse.Data[0].NextExecution).To(BeNil())
+
+							//TODO: fail here with zero timestamp
```
antoinegelloz, line 168: @gfyrag test to fix here

PR #196, opened 2023-03-31
feat: change metadata to map[string]string

gfyrag:
replace map[string]any by map[string]string on all metadata


components/ledger/pkg/machine/machine.go
```diff
@@ -30,8 +30,8 @@ func Run(m *vm.Machine, script core.RunScript) (*Result, error) {
 
 	result := Result{
 		Postings:        make([]core.Posting, len(m.Postings)),
-		Metadata:        map[string]any{},
-		AccountMetadata: map[string]core.Metadata{},
+		Metadata:        m.GetTxMetaJSON(),
+		AccountMetadata: m.GetAccountsMetaJSON(),
```
antoinegelloz, line 34: Much simpler! 🎉

components/fctl/pkg/metadata.go
```diff
@@ -21,19 +22,15 @@ func ParseMetadata(array []string) (map[string]any, error) {
 	return metadata, nil
 }
 
-func PrintMetadata(out io.Writer, metadata map[string]any) error {
+func PrintMetadata(out io.Writer, metadata map[string]string) error {
```
paul-nicolas, line 25: Why not use metadata.Metadata instead of map[string]string directly ?

gfyrag: You're right, i will change

components/ledger/pkg/storage/sqlstorage/ledger/accounts.go
```diff
@@ -25,8 +26,8 @@ const (
 type Accounts struct {
 	bun.BaseModel `bun:"accounts,alias:accounts"`
 
-	Address  string         `bun:"address,type:varchar,unique,notnull"`
-	Metadata map[string]any `bun:"metadata,type:jsonb,default:'{}'"`
+	Address  string            `bun:"address,type:varchar,unique,notnull"`
+	Metadata map[string]string `bun:"metadata,type:jsonb,default:'{}'"`
```
paul-nicolas, line 30: Should we plan a SQL migration for this ? Do we know what we currently have in the database?

gfyrag: Yes we will need a proper migration for this.

PR #207, opened 2023-04-05
feat: factorize pagination

gfyrag:


components/ledger/pkg/storage/accounts.go
```diff
@@ -0,0 +1,63 @@
+package storage
+
+import (
+	"github.com/formancehq/stack/libs/go-libs/metadata"
+)
+
+type AccountsQuery OffsetPaginatedQuery[AccountsQueryPayload]
+
+type AccountsQueryPayload struct {
+	AfterAddress    string
```
paul-nicolas, line 10: Should we add json tags on this struct since it will be the one unmarshaled from the url ?

paul-nicolas: Some for other structs

components/ledger/pkg/storage/accounts.go
```diff
@@ -0,0 +1,63 @@
...
+		Order:    OrderAsc,
+		Payload: AccountsQueryPayload{
+			Metadata: metadata.Metadata{},
+		},
+	}
+}
+
+func (a AccountsQuery) WithPageSize(pageSize uint64) AccountsQuery {
+	if pageSize != 0 {
+		a.PageSize = pageSize
```
paul-nicolas, line 29: Should we return an error when pageSize == 0 ? Maybe not here but in the caller ?

components/ledger/pkg/storage/accounts.go
gfyrag: Good point. I will add them.

components/ledger/pkg/storage/sqlstorage/ledger/accounts.go
```diff
@@ -30,58 +29,34 @@ type Accounts struct {
...
-		metadata        = p.Filters.Metadata
-		balance         = p.Filters.Balance
-		balanceOperator = p.Filters.BalanceOperator
-	)
-
-	if address != "" {
-		sb.Where("address ~* ?", "^"+address+"$")
-		t.AddressRegexpFilter = address
+
+	if p.Payload.Address != "" {
```
paul-nicolas, line 35: Payload is way less understandable than Filters, would it be possible to change it easily ?

paul-nicolas (approved):

PR #212, opened 2023-04-06
feat: wait for logs ingestion before starting

paul-nicolas:


gfyrag (approved):
Seems ok.
The initLedger MUST not be time consuming as the i think the fx context passed to the hook has a timeout.

PR #215, opened 2023-04-06
feat: Numscript: Send monetary with arithmetic

antoinegelloz:
Be able to send monetary with arithmetic computation such as:
```
vars {
  asset $asset
  monetary $mon1 = balance($account_a, $asset)
  monetary $mon2 = balance($account_b, $asset)
}
send $mon1 + $mon2 - [$asset 1] (
  source = @wallet
  destination = @account_c
)
```

### DIFF
--- a/components/ledger/pkg/machine/script/compiler/compiler.go
+++ b/components/ledger/pkg/machine/script/compiler/compiler.go
@@ -78,38 +78,58 @@ func (p *parseVisitor) VisitVariable(c parser.IVariableContext, push bool) (core
 func (p *parseVisitor) VisitExpr(c parser.IExpressionContext, push bool) (core.Type, *core.Address, *CompileError) {
 	switch c := c.(type) {
 	case *parser.ExprAddSubContext:
-		ty, _, err := p.VisitExpr(c.GetLhs(), push)
+		lhsType, lhsAddr, err := p.VisitExpr(c.GetLhs(), push)
 		if err != nil {
 			return 0, nil, err
 		}
-		if ty != core.TypeNumber {
-			return 0, nil, LogicError(c, errors.New("tried to do arithmetic with wrong type"))
-		}
-		ty, _, err = p.VisitExpr(c.GetRhs(), push)
-		if err != nil {
-			return 0, nil, err
-		}
-		if ty != core.TypeNumber {
-			return 0, nil, LogicError(c, errors.New("tried to do arithmetic with wrong type"))
-		}
-		if push {
-			switch c.GetOp().GetTokenType() {
-			case parser.NumScriptLexerOP_ADD:
-				p.AppendInstruction(program.OP_IADD)
-			case parser.NumScriptLexerOP_SUB:
-				p.AppendInstruction(program.OP_ISUB)
+		switch lhsType {
+		case core.TypeNumber:
+			rhsType, _, err := p.VisitExpr(c.GetRhs(), push)
+			if err != nil {
+				return 0, nil, err
 			}
+			if rhsType != core.TypeNumber {
+				return 0, nil, LogicError(c, fmt.Errorf(
+					"tried to do an arithmetic operation with incompatible left and right-hand side operand types: %s and %s",
+					lhsType, rhsType))
+			}
+			if push {
+				switch c.GetOp().GetTokenType() {
+				case parser.NumScriptLexerOP_ADD:
+					p.AppendInstruction(program.OP_IADD)
+				case parser.NumScriptLexerOP_SUB:
+					p.AppendInstruction(program.OP_ISUB)
+				}
+			}
+			return core.TypeNumber, nil, nil
+		case core.TypeMonetary:
+			rhsType, _, err := p.VisitExpr(c.GetRhs(), push)
+			if err != nil {
+				return 0, nil, err
+			}
+			if rhsType != core.TypeMonetary {
+				return 0, nil, LogicError(c, fmt.Errorf(
+					"tried to do an arithmetic operation with incompatible left and right-hand side operand types: %s and %s",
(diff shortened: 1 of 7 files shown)


components/ledger/pkg/machine/script/compiler/compiler.go
```diff
@@ -78,29 +79,50 @@ func (p *parseVisitor) VisitVariable(c parser.IVariableContext, push bool) (core
...
 		}
-		ty, _, err = p.VisitExpr(c.GetRhs(), push)
+
+		rightOpType, _, err := p.VisitExpr(c.GetRhs(), true)
 		if err != nil {
 			return 0, nil, err
 		}
-		if ty != core.TypeNumber {
-			return 0, nil, LogicError(c, errors.New("tried to do arithmetic with wrong type"))
+		switch rightOpType {
```
gfyrag, line 98: This does not handle monetary + number?
Is that ok?

antoinegelloz: It was not in the original design, @altitude do you think it would make sense to allow it?

components/ledger/pkg/machine/vm/program/resource.go
```diff
@@ -8,22 +8,28 @@ import (
 
 type Resource interface {
 	GetType() core.Type
+	GetMonetaryAsset() *core.Address
```
gfyrag, line 11: Adding those methods on the Resource interface feels like a code smell.
Since you have the type with GetType(), maybe you could simply cast the interface when needed?

components/ledger/pkg/machine/script/compiler/compiler.go
```diff
@@ -62,35 +68,56 @@ func (p *parseVisitor) isWorld(addr core.Address) bool {
...
-		return res.GetType(), &idx, nil
+		var assetSource *core.Address
+		if res.GetType() == core.TypeMonetary ||
+			res.GetType() == core.TypeAsset {
+			assetSource = &idx
+		}
+		return &ExprInfo{
+			typ:         res.GetType(),
+			addr:        &idx,
+			assetSource: assetSource,
```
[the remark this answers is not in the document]
antoinegelloz: I need to store the address of the asset only when the variable is of type `asset` or `monetary`

components/ledger/pkg/machine/script/compiler/compiler.go
```diff
@@ -100,159 +127,292 @@ func (p *parseVisitor) VisitExpr(c parser.IExpressionContext, push bool) (core.T
...
+		accounts, _, _, compErr := p.VisitSource(c.Source(), func() {
+			p.PushAddress(*assetInfo.assetSource)
+		}, true)
 		if compErr != nil {
 			return compErr
 		}
-		if typ != core.TypeAsset {
-			return LogicError(c, fmt.Errorf(
-				"send monetary all: the expression should be of type 'asset' instead of '%s'", typ))
+		for accAddr := range accounts {
```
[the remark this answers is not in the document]
antoinegelloz: ok!

components/ledger/pkg/machine/vm/program/resource.go
antoinegelloz: I agree that this is not ideal, but I can't cast the interface with only the type, because a `core.TypeMonetary` can be of concrete type `program.Monetary` or `program.VariableAccountBalance`.

gfyrag (approved):

components/ledger/pkg/machine/script/compiler/compiler.go
```diff
@@ -78,38 +78,58 @@ func (p *parseVisitor) VisitVariable(c parser.IVariableContext, push bool) (core
...
+			}
+			if push {
+				switch c.GetOp().GetTokenType() {
+				case parser.NumScriptLexerOP_ADD:
+					p.AppendInstruction(program.OP_MONETARY_ADD)
+				case parser.NumScriptLexerOP_SUB:
+					p.AppendInstruction(program.OP_MONETARY_SUB)
+				}
+			}
+			return core.TypeMonetary, lhsAddr, nil
```
Azorlogh, line 123: I think this is misleading. I know we use the left monetary's address to gain access to its asset, but it isn't clearly expressed in this API (it looks like the second return value will be the current expr's value even though it isn't).  This is why in my draft I made a separate struct for VisitExpr's return, so that it's clearer that `asset_source` is *not* the actual address (though asset_source isn't a very good name because of the other meaning of "source" 😅)

components/ledger/pkg/machine/script/compiler/compiler.go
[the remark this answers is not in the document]
Azorlogh: Ah just saw this, I mentioned [in this comment](https://github.com/formancehq/stack/pull/215#discussion_r1169299821) why I added it, it's because when doing `$mon_a + $mon_b`, there is no resource address (the expression's result is not a resource), however we still need the corresponding asset for the `send` mechanism, so as a hack we get the `assetSource` of one of the operands.
Maybe it's not too big of a deal, but I'm afraid that if we use the `addr` field for this, in the future we might accidentally end up pushing the `addr` thinking that we just pushed the expression's result, and it will silently push the wrong value without complaining.

Azorlogh (commented):
Apart from my other comment, the rest LGTM :+1: 

Just a side note, I think `program.Monetary` could just be replaced by `OP_MONETARY_NEW` (we could just say that whenever we write [EUR/2 42], it creates 2 resources EUR/2 and 42, and the monetary is just created from OP_MONETRARY_NEW.  
But maybe I'm missing some details. In any case it's somewhat unrelated to this PR.

paul-nicolas (approved):

PR #216, opened 2023-04-06
feat: add cache eviction

gfyrag:


components/ledger/pkg/storage/sqlstorage/cli.go
```diff
@@ -19,6 +19,9 @@ const (
 	StoragePostgresConnectionStringFlag = "storage-postgres-conn-string"
 )
 
+// TODO(gfyrag): maybe move flag handling inside cmd/internal (as telemetry flags)
```
paul-nicolas, line 22: True, we should discuss about a convention about flags

paul-nicolas (approved):

PR #217, opened 2023-04-07
fix: concurrent reverts

gfyrag:


components/ledger/pkg/api/apierrors/errors.go
```diff
@@ -64,7 +65,9 @@ func coreErrorToErrorCode(err error) (int, string, string) {
 	case
 		ledger.IsValidationError(err),
 		state.IsPastTransactionError(err),
-		runner.IsNoPostingsError(err):
+		runner.IsNoPostingsError(err),
+		errors.Is(err, revert.ErrAlreadyReverted),
```
paul-nicolas, line 69: In order to keep the same syntax, should we create function ? (not a big deal for me, but just asking !)

components/ledger/pkg/ledger/revert/reverter.go
```diff
@@ -0,0 +1,101 @@
...
+
+type Reverter struct {
+	sync.Map
+	store       Store
+	runner      Runner
+	logIngester LogIngester
+}
+
+func (r *Reverter) RevertTransaction(ctx context.Context, id string) (*core.ExpandedTransaction, error) {
+	_, loaded := r.Map.LoadOrStore(id, struct{}{})
```
paul-nicolas, line 45: We really have to be sure that there is one ledger instance !

components/ledger/pkg/api/apierrors/errors.go
gfyrag: yes, agreed

components/ledger/pkg/ledger/revert/reverter.go
gfyrag: Sorry, I don't understand ^^

paul-nicolas (approved):

PR #220, opened 2023-04-11
fix: mapping ignore destinations accounts

gfyrag:


components/ledger/pkg/ledger/execute_txsdata.go
```diff
@@ -114,6 +114,9 @@ func (l *Ledger) ExecuteTxsData(ctx context.Context, preview bool, txsData ...co
 				if account == core.WORLD {
 					continue
 				}
+				if volume.Balance().Gt(txVolumeAggr.PreCommitVolumes[account][asset].Balance()) {
```
antoinegelloz, line 117:
```suggestion
				if volume.Balance().Gte(txVolumeAggr.PreCommitVolumes[account][asset].Balance()) {
```

It should be Gte, otherwise postings with amount 0 will fail.

PR #224, opened 2023-04-11
feat: add some benchmark on ledger

gfyrag:


components/ledger/pkg/ledger/query/worker.go
```diff
@@ -424,11 +370,13 @@ func buildData(
 	return logsData, nil
 }
 
-func (w *Worker) QueueLog(log *core.LogHolder) {
+func (w *Worker) QueueLog(ctx context.Context, log *core.LogHolder) error {
 	select {
-	case <-w.ctx.Done():
+	case <-w.stoppedChan:
```
paul-nicolas, line 375: Should we wait on the ctx here too ?

components/ledger/pkg/storage/sqlstorage/ledger/store.go
```diff
@@ -106,10 +106,7 @@ func (s *Store) RunInTransaction(ctx context.Context, f func(ctx context.Context
 	newStore.isInitialized = s.isInitialized
 
 	defer func() {
-		// TODO(gfyrag): add recover() to rollback txs
-		if err != nil {
-			_ = tx.Rollback()
-		}
+		_ = tx.Rollback()
```
paul-nicolas, line 109: We rollback everytime ?

paul-nicolas: Why removing the error check ?

gfyrag: The error check was not ok.
If the 'tx.Commit()' return an error, the rollback is not called.
The error check is not needed, as calling tx.Rollback() will have no effect if the transaction is already committed.

components/ledger/pkg/ledger/query/worker.go
gfyrag: Good question.
We need to NOT cancel the log ingestion if the request is cancelled, as, at this moment, the cache is already filled with new information.

gfyrag: But we should be able to properly cancel the request and release any potential used resources.

paul-nicolas: Good point about the cache !

gfyrag: If we update the cache after the log ingestion, we will lock account until the log ingestion.
It would cancel the benefits of the CRQS no?

paul-nicolas (approved):

PR #235, opened 2023-04-14
feat: allow core to handle transactions aynchronously

gfyrag:


paul-nicolas (approved):
Nice, we will just have to plug it into the API in the future ! Nice one !

PR #243, opened 2023-04-17
feat: add ballast package

paul-nicolas:
cf: https://blog.twitch.tv/en/2019/04/10/go-memory-ballast-how-i-learnt-to-stop-worrying-and-love-the-heap/

Could be use by some of our biggest clients

### DIFF
--- a/components/ledger/cmd/serve.go
+++ b/components/ledger/cmd/serve.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"github.com/formancehq/ledger/pkg/ledger/cache"
+	"github.com/formancehq/stack/libs/go-libs/ballast"
 	"github.com/formancehq/stack/libs/go-libs/httpserver"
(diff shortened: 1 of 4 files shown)


components/ledger/cmd/serve.go
```diff
@@ -14,6 +15,8 @@ const (
 	cacheEvictionPeriodFlag  = "cache-eviction-period"
 	cacheEvictionRetainDelay = "cache-eviction-retain-delay"
 	queryLimitReadLogsFlag   = "query-limit-read-logs"
+
+	ballastSizeInBytesFlag = "ballast-size-in-bytes"
```
gfyrag, line 19: "-in-bytes" is maybe useless.
I think this should be mentioned on usage but not required in the flag.

paul-nicolas: Done thanks !

gfyrag (approved):

PR #245, opened 2023-04-17
chore: Ledger: deprecation message for sqlite

antoinegelloz:


components/ledger/cmd/root.go
```diff
@@ -59,6 +62,9 @@ func NewRootCommand() *cobra.Command {
 		Short:             "Numary",
 		DisableAutoGenTag: true,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			if viper.GetString(storageDriverFlag) == "sqlite" {
+				fmt.Printf("WARNING: SQLite is being deprecated and will not be supported starting from the v2 of the Formance Ledger. Please use Postgres instead.\n")
```
gfyrag, line 66: Maybe you would write on stderr instead of stdout?
From an ops perspective, we can easily discard stdout but not stderr.

paul-nicolas (approved):

gfyrag (approved):

PR #247, opened 2023-04-18
feat: use logger with json formatter

paul-nicolas:
Asked by @flemzord


tests/integration/internal/bootstrap.go
```diff
@@ -80,7 +80,7 @@ var _ = BeforeEach(func() {
 	l := logrus.New()
 	l.Out = GinkgoWriter
 	l.Level = logrus.DebugLevel
-	ctx = logging.ContextWithLogger(ctx, logging.NewLogrus(l))
+	ctx = logging.ContextWithLogger(ctx, logging.NewLogrusWithJsonFormatter(l))
```
gfyrag, line 83: Not required here i think. That's integration testing, json is not easy to read ^^

gfyrag (approved):

PR #258, opened 2023-04-20
feat: add index on postings

gfyrag:


components/ledger/pkg/storage/sqlstorage/ledger/migrates/0-init-schema/postgres.sql
```diff
@@ -82,7 +82,8 @@ CREATE TABLE IF NOT EXISTS "VAR_LEDGER_NAME".postings (
     amount bigint not null,
     asset varchar not null,
     source jsonb not null,
-    destination jsonb not null
+    destination jsonb not null,
+    index int8
```
paul-nicolas, line 86: Maybe a primary key ?

PR #272, opened 2023-04-23
opti: remove sql transactions

gfyrag:


components/ledger/pkg/ledger/query/worker.go
```diff
@@ -101,7 +101,7 @@ l:
 						return nil
 
 					case w.jobs <- w.pending:
-						w.pending = make([]*core.LogHolder, 0)
+						w.pending = make([]*core.LogHolder, 0, 1024)
```
paul-nicolas, line 104: Should we make this configurable ?

components/ledger/pkg/storage/sqlstorage/ledger/migrates/0-init-schema/postgres.sql
```diff
@@ -95,8 +91,9 @@ CREATE TABLE IF NOT EXISTS "VAR_LEDGER_NAME_v2_0_0".volumes (
     asset character varying not null,
     input numeric not null,
     output numeric not null,
+    more_recent_log_id bigint references "VAR_LEDGER_NAME_v2_0_0".logs_v2,
 
-    unique(account, asset)
+    primary key (account, asset)
```
paul-nicolas, line 96: Nice for all primary keys !

components/ledger/pkg/storage/sqlstorage/ledger/migrates/0-init-schema/postgres.sql
```diff
@@ -49,14 +47,12 @@ CREATE TABLE IF NOT EXISTS "VAR_LEDGER_NAME_v2_0_0".logs_ingestion (
 
 --statement
 CREATE TABLE IF NOT EXISTS "VAR_LEDGER_NAME_v2_0_0".logs_v2 (
-    id bigint,
+    id bigint primary key,
```
paul-nicolas, line 50: I think we should add them in bun models too, for example:

ID             uint64    `bun:"id,unique,type:bigint"` ->
I think we should add them in bun models too, for example:

ID             uint64    `bun:"id,unique,pk,type:bigint"`

paul-nicolas (approved):

PR #274, opened 2023-04-24
feat: remove tracer from health road

paul-nicolas:


components/payments/internal/app/api/router.go
```diff
@@ -31,10 +27,15 @@ func httpRouter(logger logging.Logger, store *storage.Storage, serviceInfo api.S
 	})
 
 	rootMux.Path("/_health").Handler(healthHandler(store))
-	rootMux.Path("/_live").Handler(liveHandler())
-	rootMux.Path("/_info").Handler(api.InfoHandler(serviceInfo))
 
-	authGroup := rootMux.Name("authenticated").Subrouter()
+	subRouter := rootMux.NewRoute().Subrouter()
+	if viper.GetBool(otelTracesFlag) {
+		subRouter.Use(otelmux.Middleware(serviceName))
```
gfyrag, line 33: Put there, i think the middleware will no catch errors catched by the recoveryHandler. Can you check if otelmux catch panics?

paul-nicolas: You're right, let me think about something to fix that !

gfyrag (approved):

PR #303, opened 2023-05-11
refactor: cqrs/memory management

gfyrag:


components/ledger/.gitignore
```diff
@@ -15,5 +15,4 @@ ledger.test
 antlr-*-complete.jar
 go.work
 go.work.sum
-ledger
```
paul-nicolas, line 18: Why did you remove that ? it should prevent us to push the binary, am I right ?

gfyrag: yes, but it was hide the "ledger" package ^^
This has probably been added by error.

paul-nicolas (approved):

PR #317, opened 2023-05-17
feat: add logs for stargate + fixes for client/server

paul-nicolas:


flemzord (approved):

components/stargate/cmd/client.go
```diff
@@ -35,6 +35,10 @@ const (
 	StargateAuthClientIDFlag                     = "stargate-auth-client-id"
 	StargateAuthClientSecretFlag                 = "stargate-auth-client-secret"
 	StargateAuthIssuerURLFlag                    = "stargate-auth-issuer-url"
+	StargateAuthMaxRetriesFlag                   = "stargate-auth-max-retries"
```
gfyrag, line 38: Why retry authentication?
I think we should let the underlying system (k8s in our context), retry if necessary.

PR #336, opened 2023-05-25
feat: add capability to restore stacks using fctl

gfyrag:
- feat: add capability to disable a stack
- feat: add stack restore


components/fctl/membershipclient/model_billing_portal.go
```diff
@@ -114,3 +114,5 @@ func (v *NullableBillingPortal) UnmarshalJSON(src []byte) error {
 	v.isSet = true
 	return json.Unmarshal(src, &v.value)
 }
+
```
paul-nicolas, line 117: Is this normal to have all files with 2 more lines at the end?

PR #365, opened 2023-06-15
feat: improve accounts/balances/volumes queries

paul-nicolas:


components/ledger/pkg/storage/ledgerstore/accounts.go
```diff
@@ -21,18 +22,35 @@ const (
...
+var accountNameRegex = regexp.MustCompile(`^[a-zA-Z_0-9]+$`)
+
 type Accounts struct {
 	bun.BaseModel `bun:"accounts,alias:accounts"`
 
-	Address  string            `bun:"address,type:varchar,unique,notnull"`
-	Metadata map[string]string `bun:"metadata,type:jsonb,default:'{}'"`
+	Address     string            `bun:"address,type:varchar,unique,notnull"`
+	Metadata    map[string]string `bun:"metadata,type:jsonb,default:'{}'"`
+	AddressJson []string          `bun:"address_json,type:jsonb"`
```
gfyrag, line 35: should be account_json no?

paul-nicolas: No, it's address_json to be in accordance with the address field !

PR #366, opened 2023-06-19
feat: convert all webhook command to json

Dav-14:


components/fctl/cmd/webhooks/activate.go
```diff
@@ -10,55 +10,73 @@ import (
 	"github.com/spf13/cobra"
 )
 
+func RunActivateWebhookfunc(cmd *cobra.Command, args []string) error {
```
gfyrag, line 13: "func" prefix is not idomatic in go.

PR #380, opened 2023-06-22
feat: Add Helm for deploy regions

flemzord:


components/agent/helm/values.yaml
```diff
@@ -45,21 +45,18 @@ affinity: {}
...
   authentication:
-    mode: token
-    token: "sdfsdf"
-    # OAuth2 mode
-    # mode: bearer
-    # issuer: https://app.formance.cloud
-    # clientSecret: YYY
+    mode: bearer
+    issuer: https://app.formance.cloud/api
+    clientID: ""
```
gfyrag, line 61: no clientID here, the client is is the id of the agent

gfyrag (approved):

Issue #385, opened 2023-06-22
breaking changes in garden v0.13.3


kislow:
While trying to deploy the formance-operator using [garden deploy](https://github.com/formancehq/stack/blob/main/components/operator/README.md#running-on-the-cluster) the following error messages occur:

```sh
# new garden version needs backwards compatibility or update config files accordingly 
"apiVersion" is missing in the Project config. Assuming "garden.io/v0" for backwards compatibility with 0.12. 
The "apiVersion"-field is mandatory when using the new action Kind-configs. 
A detailed migration guide is available at https://docs.garden.io/tutorials/migrating-to-bonsai

# providers needs to be removed from environments
Error validating environment default: key "providers" is not allowed at path [providers]. 
Available keys: name, defaultNamespace, production, varfile, variables)
```

PR #386, opened 2023-06-22
fix: garden/deployments

kislow:
Fix issue #385

- Add backwards compatibility
- update project config (remove providers key in environments)


components/operator/project.garden.yaml
```diff
@@ -7,12 +8,10 @@ variables:
 
 environments:
   - name: default
-    providers:
-    - name: kubernetes
-      setupIngressController: false
 
 providers:
   - name: kubernetes
+    setupIngressController: false
```
gfyrag, line 14: Can you use a variable please? To avoid breaking changes for those who use the default configuration. With an override for your use case inside a `garden.default.env`, it should be ok for your use case.

kislow: For my use case I don't need it and would've remove it. Thought it might be a good idea to keep it in, because the current config file contained this option. Do you still want me to use a variable or can we simply remove it?

gfyrag: Ok, got it! Thks.

gfyrag (approved):

paul-nicolas (approved):

PR #412, opened 2023-07-04
feat: add balance in WalletSubject Struct

Dav-14:


Dav-14:
feat: feat: add balance omit empty on Subject

PR #414, opened 2023-07-04
fix: metadata printing between ledger v1 et v2

Dav-14:


components/fctl/cmd/ledger/internal/transaction.go
```diff
@@ -167,7 +189,7 @@ func CreateTransaction(client *formance.Formance, ctx context.Context, request o
 		if err != nil {
 			return nil, err
 		}
-
+		fmt.Println("eazeaz", v.CreateTransactionResponse)
```
paul-nicolas, line 192: Maybe we should remoe this print ?

paul-nicolas (approved):

PR #423, opened 2023-07-10
feat: handle not found errors from payments storage

paul-nicolas:
Fixes NUM-1764


flemzord (approved):

components/payments/internal/app/api/module.go
```diff
@@ -88,6 +98,18 @@ func handleServerError(w http.ResponseWriter, r *http.Request, err error) {
 	}
 }
 
+func handleNotFoundError(w http.ResponseWriter, r *http.Request, err error) {
```
gfyrag, line 101: There is a common helper in the shared libs, maybe use it?

gfyrag (approved):

components/payments/internal/app/api/module.go
paul-nicolas: Done !

PR #424, opened 2023-07-10
feat: add pong on grpc messages

gfyrag:


components/agent/server.proto
```diff
@@ -6,7 +6,7 @@ package server;
 
 service Server {
   rpc Connect(ConnectRequest) returns (stream Order) {}
-  rpc Join(stream StatusChanged) returns (stream Order) {}
+  rpc Join(stream Message) returns (stream Order) {}
```
paul-nicolas, line 9: We will have to update both membership-api and all agents together in order to not break everything

paul-nicolas (approved):

PR #436, opened 2023-07-18
feat: remove cobra dependency in Controller interface

Dav-14:
FEAT ENG-118

Every file is structured that way:
1- Store
2- Config
3- Controller
4- Cobra Command

### DIFF
--- a/components/fctl/cmd/auth/clients/create.go
+++ b/components/fctl/cmd/auth/clients/create.go
@@ -1,13 +1,22 @@
 package clients
 
 import (
+	"flag"
 	"fmt"
 	"strings"
+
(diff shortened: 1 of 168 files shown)


components/fctl/cmd/auth/clients/create.go
```diff
@@ -22,90 +31,96 @@ type CreateClient struct {
...
+	flags.String(internal.PostLogoutRedirectUriFlag, "", "Post logout redirect uris") // Slice
+
+	return fctl.NewControllerConfig(
+		useCreate,
+		descriptionCreate,
+		shortCreate,
+		[]string{
+			"setup", "set",
+		},
+		os.Stdout,
```
gfyrag, line 54: use cmd.Output()

PR #437, opened 2023-07-18
feat: upgrade sdk to use bigints

gfyrag:


paul-nicolas (approved):
Just missing the integration tests fix with big ints, but otherwise, lgtm !

PR #449, opened 2023-07-21
fix: 500 status code

gfyrag:


devbox.lock
```diff
@@ -42,4 +42,4 @@
       "version": "1.22.19"
     }
   }
-}
\ No newline at end of file
+}
```
paul-nicolas, line 45: Be careful the ci will fail when doing a release with this change

paul-nicolas (approved):

paul-nicolas (approved):

PR #470, opened 2023-08-07
feat: upgrade garden project

Dav-14:
- Bump garden project version to garden.io/v1
- It allow us to trigger action https://docs.garden.io/reference/action-types
- `kind: Module` is deprecated
- Convert every `kind: Module` to `kind: Deploy`
- Install default registry as k3d-registry.host.k3d.internal on static port 12345
- Update makefile:
  - docker-push
  - docker-build
  - helm-local-install
  - helm-local-upgrade


components/operator/garden/nats/garden.yaml
```diff
@@ -1,16 +1,17 @@
 ---
-apiVersion: garden.io/v0
```
gfyrag, line 2: maybe keep the version field on v1?

components/operator/garden/otel-collector/garden.yaml
```diff
@@ -17,3 +15,27 @@ services:
       --set config.exporters.jaeger.endpoint=jaeger:14250
       --set config.service.pipelines.traces.exporters[0]=jaeger
       --set config.exporters.jaeger.tls.insecure=true
+---
```
gfyrag, line 18: You can remove commented sections (also on other files)

components/operator/garden/nats/garden.yaml
Dav-14: I've added back, the apiVersion description, to all Build, Deploy actions

PR #479, opened 2023-08-11
feat: disable stack

Dav-14:
Related to membership, `feat/disable-stack`

- [X] Update membership open api client, 
- [ ] Add status for stack disabled_at != null


Dav-14:
Fixes ENG-104

PR #496, opened 2023-08-23
feat: make cpu and memory resources requests configurable in operator

paul-nicolas:


Dav-14 (approved):

components/operator/apis/stack/v1beta3/common.go
Dav-14, line 1: resources => resource ?

Each resources request & limit is reference by the parent key `resources` in kubernetes, would it be nice to keep the same syntax on the Struct ?
ResourceProperties => Resources

components/operator/internal/handlers/handler_webhooks.go
Dav-14, line 1: - It migth be a good idea to put default resources in the Helm chart directly and override at deployment time ?

gfyrag (approved):

gfyrag (approved):

PR #558, opened 2023-10-02
feat: add deployment strategy on ledger

gfyrag:
Add deployment strategy for the ledger on the operator.
The operator now deploy two Deployments for the ledger : 
* A writer : Numbers of replicas to 1
* Readers: Number of replicas initially set to 1 but can be changed

Also, the gateway will route all http requests with GET, OPTIONS or HEAD methods to the readers, others to the writer.

### DIFF
--- a/components/operator/internal/modules/ledger/handler.go
+++ b/components/operator/internal/modules/ledger/handler.go
@@ -1,6 +1,7 @@
 package ledger
 
 import (
+	"github.com/formancehq/stack/libs/go-libs/pointer"
 	"net/http"
 	"strconv"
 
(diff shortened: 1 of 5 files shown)


paul-nicolas (approved):

components/operator/internal/modules/ledger/handler.go
```diff
@@ -100,8 +98,35 @@ func (l module) Versions() map[string]modules.Version {
...
+					HasVersionEndpoint:      true,
+					ExposeHTTP:              modules.DefaultExposeHTTP,
+					NeedTopic:               true,
+					Container:               createContainer(modules.NewEnv()),
+				}
+				ret := modules.Services{&main}
+
+				if ctx.Configuration.Spec.Services.Ledger.DeploymentStrategy == v1beta3.DeploymentStrategyMonoWriterMultipleReader {
+					cp := main
+					cp.Name = "read"
```
flemzord, line 116: write ?

PR #559, opened 2023-10-03
feat: agent version on connection metadata

gfyrag:


Dav-14 (approved):

components/ledger/libs/pgtesting/postgres.go
```diff
@@ -253,5 +253,5 @@ func CreatePostgresServer(opts ...option) error {
 		return nil
 	}
 
-	return errors.New("timeout waiting for server ready")
+	return errors.New("timeout waiting for server ready 2")
```
Dav-14, line 256: why ready 2 ?

PR #613, opened 2023-10-10
feat: remove stakater reloader usage

gfyrag:
Fixes ENG-91


Dav-14 (approved):

components/operator/internal/modules/module.go
```diff
@@ -114,26 +147,19 @@ func (r *moduleReconciler) installModule(ctx context.Context, registeredModules
...
-	for _, version := range sortedVersions(r.module) {
-		if !r.Versions.IsHigherOrEqual(r.module.Name(), version) {
-			break
-		}
-		chosenVersion = r.module.Versions()[version]
-		chosenVersionLabel = version
-		if chosenVersion.PreUpgrade == nil {
-			continue
+	for _, version := range sortedVersions {
+		if chosenVersion.PreUpgrade != nil {
```
paul-nicolas, line 151: version instead of chosenVersion ?

paul-nicolas: No more check if IsHigherOrEqual here also ?

gfyrag: The check is made earlier. The 
```
if version == chosenVersionLabel {
			break
		}
```
prevent to run undesirable migration.
BUt you're right for chosenVersion / version. I change.

paul-nicolas (approved):

PR #654, opened 2023-10-23
feat(payments): split payments into two services

paul-nicolas:
Fixes ENG-136


components/payments/cmd/api/internal/api/module.go
```diff
@@ -0,0 +1,141 @@
...
+	if errors.Is(err, storage.ErrDuplicateKeyValue) {
+		handleErrorBadRequest(w, r, errors.New("unique reference already existing"))
+	} else if errors.Is(err, storage.ErrNotFound) {
+		handleNotFoundError(w, r, err)
+	} else {
+		handleServerError(w, r, err)
+	}
+}
+
+func handleErrorBadRequest(w http.ResponseWriter, r *http.Request, err error) {
```
gfyrag, line 72: Maybe use shared lib for all these response?

gfyrag (approved):

components/payments/cmd/api/internal/api/module.go
paul-nicolas: Done thanks !

PR #689, opened 2023-10-27
feat: display effective status, and state in 'status'

Dav-14:


components/fctl/membership-swagger.yaml
```diff
@@ -197,15 +204,9 @@ paths:
         schema:
           type: string
         required: true
-      - name: deleted
```
gfyrag, line 200: Doing this case a breaking change for fctl

Dav-14: same for `--disabled` 


1/ Do I mark them deprecated ?
2/ Do I keep those query paramaters membership side ? (Map it on --all)

components/fctl/cmd/stack/show.go
```diff
@@ -107,11 +108,18 @@ func (c *StackShowController) Run(cmd *cobra.Command, args []string) (fctl.Rende
 
 	stackClient, err := fctl.NewStackClient(cmd, cfg, stack)
 	if err != nil {
+		if strings.Contains(err.Error(), "404 page not found") && stack.Status != "ACTIVE" {
```
gfyrag, line 111: Beurk :)

gfyrag (approved):

gfyrag (approved):
Fix the 404 please ;)

PR #783, opened 2023-10-31
feat(payments): insert all data in search via benthos

paul-nicolas:
Fixes ENG-148


components/operator/internal/controllers/stack/testdata/monopod-disabled-one-service/results/configmaps--v1/search-benthos-streams.yaml
```diff
@@ -299,50 +299,70 @@ data:
...
-                            "initialAmount": this.payload.initialAmount,
-                            "createdAt": this.payload.createdAt
-                        },
-                        "kind": "PAYMENT",
-                        "when": this.date
+                      "query": {
+                        "bool": {
+                          "must": [
+                            {
+                              "bool": {
```
gfyrag, line 320: What is the behavior if matching against multiple value of "kind" ?

components/search/pkg/searchhttp/http.go
```diff
@@ -128,6 +128,12 @@ func resolveQuery(r *http.Request) (*cursorTokenInfo, BaseQuery, error) {
 				qq.WithSort("txid", esquery.OrderDesc)
 			case "PAYMENT":
 				qq.WithSort("reference", esquery.OrderDesc)
+			case "PAYMENT_TRANSFER_INITIATION":
```
gfyrag, line 131: I would remove this code. 
but it is ok for now.

gfyrag (approved):

components/operator/internal/controllers/stack/testdata/monopod-disabled-one-service/results/configmaps--v1/search-benthos-streams.yaml
paul-nicolas: Yes, must = AND and should = OR, so in this case, everything works as expected !

PR #788, opened 2023-11-03
feat: add aggregation using metadata

gfyrag:
Fixes ENG-204

### DIFF
--- a/components/ledger/internal/storage/ledgerstore/balances.go
+++ b/components/ledger/internal/storage/ledgerstore/balances.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/formancehq/ledger/internal/storage"
 	"math/big"
 
 	ledger "github.com/formancehq/ledger/internal"
@@ -18,7 +19,7 @@ func (store *Store) GetAggregatedBalances(ctx context.Context, q *GetAggregatedB
 	type Temp struct {
 		Aggregated ledger.VolumesByAssets `bun:"aggregated,type:jsonb"`
 	}
-	return fetchAndMap[*Temp, ledger.BalancesByAssets](store, ctx,
+	ret, err := fetchAndMap[*Temp, ledger.BalancesByAssets](store, ctx,
 		func(temp *Temp) ledger.BalancesByAssets {
 			return temp.Aggregated.Balances()
 		},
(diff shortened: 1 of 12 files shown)


components/ledger/internal/storage/ledgerstore/balances.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/formancehq/ledger/internal/storage"
```
paul-nicolas, line 7: maybe group this import with the others below ?

flemzord (approved):

components/ledger/internal/storage/ledgerstore/balances.go
```diff
@@ -27,6 +28,7 @@ func (store *Store) GetAggregatedBalances(ctx context.Context, q *GetAggregatedB
 				NewSelect().
 				Table(MovesTableName).
 				ColumnExpr("distinct on (moves.account_address, moves.asset) moves.*").
+				Join("left join accounts_metadata am on am.address = moves.account_address").
```
paul-nicolas, line 31: Maybe it's not possible in the code, but can we join only if we aggregate using metadata instead of all queries ?

paul-nicolas (approved):

components/ledger/internal/storage/ledgerstore/balances.go
gfyrag: It should be possible.

components/ledger/internal/storage/ledgerstore/balances.go
```diff
@@ -31,6 +32,7 @@ func (store *Store) GetAggregatedBalances(ctx context.Context, q *GetAggregatedB
 				Apply(filterPIT(q.Options.Options.PIT, "insertion_date")) // todo(gfyrag): expose capability to use effective_date
 
 			if q.Options.QueryBuilder != nil {
+				joinOnMetadataAdded := false
```
paul-nicolas, line 35: Never putted to true ?

components/ledger/internal/storage/ledgerstore/balances.go
```diff
@@ -44,6 +46,18 @@ func (store *Store) GetAggregatedBalances(ctx context.Context, q *GetAggregatedB
 						default:
 							return "", nil, fmt.Errorf("unexpected type %T for column 'address'", address)
 						}
+					case metadataRegex.Match([]byte(key)):
+						if operator != "$match" {
+							return "", nil, errors.New("'metadata' column can only be used with $match")
+						}
+						match := metadataRegex.FindAllStringSubmatch(key, 3)
+						if !joinOnMetadataAdded {
```
paul-nicolas, line 54: If joinOnMetadataAdded instead of if !joinOnMetadataAdded if I understand the name of the var ?

paul-nicolas (approved):

PR #835, opened 2023-11-08
feat: add plugin audit

Dav-14:


components/gateway/Caddyfile
```diff
@@ -101,13 +105,33 @@
 	}
 }
 
+(audit) {
```
paul-nicolas, line 108: Should we add it also for /versions and all other api calls returning 502 ?

components/gateway/pkg/plugins/audit.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+
+	return conn, nil
+}
+
+func newNatsPublisherWithConn(conn *nats.Conn, logger watermill.LoggerAdapter, config wNats.PublisherConfig) (*wNats.Publisher, error) {
+	return wNats.NewPublisherWithNatsConn(conn, config.GetPublisherPublishConfig(), logger)
+}
+
+func (a *Audit) provisionNatsPublisher() error {
```
paul-nicolas, line 202: Should those functions about instantiating NATS be in go-libs ? (Just asking for opinions here, I'm fine with either letting this code here or putting it in golibs)

components/gateway/pkg/plugins/audit.go
```diff
@@ -0,0 +1,394 @@
...
+	buf.Reset()
+	defer a.bufPool.Put(buf)
+
+	rww := NewResponseWriterWrapper(w, buf)
+	if err := next.ServeHTTP(rww, r); err != nil {
+		return err
+	}
+
+	response := messages.NewHttpResponse(
+		*rww.statusCode,
```
paul-nicolas, line 330: I have a doubt here, should we check that status code is never nil ?

paul-nicolas: Nevermind, there is a default inside the wrapper

paul-nicolas (approved):

components/gateway/Caddyfile
gfyrag: I think we should log every requests.
@flemzord ?

components/gateway/pkg/plugins/audit.go
```diff
@@ -0,0 +1,394 @@
...
+
+	return res, nil
+}
+
+// Implements the caddy.Provisioner interface.
+func (a *Audit) Provision(ctx caddy.Context) error {
+	// We have to instantiate a different logger here instead of using
+	// caddy's logger because we need a different interface for the watermill
+	// logger.
+	a.logger = logging.NewLogrus(logrus.New())
```
gfyrag, line 169: Can you implement the watermil logger interface using the caddy logger as underlying logger?

components/gateway/pkg/plugins/audit.go
gfyrag: Yes, it smells like copy/paste no?

components/gateway/pkg/plugins/audit.go
```diff
@@ -0,0 +1,394 @@
...
+		Body:   "",
+	}
+
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		if !errors.Is(err, io.EOF) {
+			return err
+		}
+	}
+
```
gfyrag, line 313: You need to close the original body before replacing it.

components/gateway/Caddyfile
flemzord: All calls must be logged

paul-nicolas (approved):

gfyrag (approved):

PR #836, opened 2023-11-09
feat: refactor error handling and add tests

gfyrag:
Fixes ENG-9


flemzord (approved):

components/ledger/internal/api/v1/controllers_accounts.go
```diff
@@ -7,19 +7,17 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/formancehq/ledger/internal/api/shared"
+	"github.com/formancehq/ledger/internal/api/backend"
```
paul-nicolas, line 10: I have a global question here: should we add a lint in order to group imports ? (just my own opinion on it, feel free to tell me what you think)

components/ledger/internal/api/v1/controllers_accounts.go
```diff
@@ -133,48 +130,46 @@ func getAccount(w http.ResponseWriter, r *http.Request) {
 
 	acc, err := l.GetAccountWithVolumes(r.Context(), query)
 	if err != nil {
-		ResponseError(w, r, err)
+		sharedapi.InternalServerError(w, r, err)
```
paul-nicolas, line 133: If the account does not exist, should we send a NotFound error instead?

components/ledger/internal/api/v2/controllers_accounts.go
```diff
@@ -80,55 +86,53 @@ func getAccount(w http.ResponseWriter, r *http.Request) {
...
-		sharedapi.BadRequest(w, shared.ErrValidation, err)
+		sharedapi.BadRequest(w, ErrValidation, err)
 		return
 	}
 	query.PITFilter = *pitFilter
 
 	acc, err := l.GetAccountWithVolumes(r.Context(), query)
 	if err != nil {
-		shared.ResponseError(w, r, err)
+		sharedapi.InternalServerError(w, r, err)
```
paul-nicolas, line 96: Same here s hould we send a not found error ?

components/ledger/internal/api/v1/controllers_accounts.go
gfyrag: Account always exists ;)

components/ledger/internal/api/v2/controllers_accounts.go
gfyrag: Same as above.

components/ledger/internal/api/v1/controllers_accounts.go
gfyrag: We already use the goimports linter on the main .golangci.yml

paul-nicolas (approved):

PR #848, opened 2023-11-13
feat(ledger): add 'ledger' column on all tables

gfyrag:
Not optimized, the 'ledger' column will be included in a partition in a later PR.

Fixes ENG-227


components/ledger/libs/go.mod
```diff
@@ -1,3 +1,161 @@
 module github.com/formancehq/stack/libs/go-libs
 
 go 1.19
+
```
paul-nicolas, line 4: Is it restored also in libs/go-libs ?

paul-nicolas (approved):

components/ledger/libs/go.mod
paul-nicolas: Already done on main sorry about this comment

flemzord (approved):

PR #862, opened 2023-11-15
feat(ledger): make ledger creation manual

gfyrag:
Fixes ENG-235


components/ledger/internal/api/v2/routes.go
```diff
@@ -37,7 +37,8 @@ func NewRouter(
 		router.Get("/_info", getInfo(b))
 
 		router.Route("/{ledger}", func(router chi.Router) {
-			router.Post("/", configureLedger(b))
+			router.Post("/", createLedger(b))
+			router.Get("/", getLedger(b))
```
paul-nicolas, line 41: Do we want a list also to display them on console ?

components/ledger/internal/engine/resolver.go
```diff
@@ -67,6 +69,15 @@ func NewResolver(storageDriver *driver.Driver, options ...option) *Resolver {
 	return r
 }
 
+func (r *Resolver) startLedgerUsingStore(ctx context.Context, name string, store *ledgerstore.Store) *Ledger {
```
paul-nicolas, line 72: start is called in the getLedger and also in the create ? Should we have a way to know if the ledger is already started or not in order to not launch the same goroutine ?

paul-nicolas: My bad, it's not launch in the get since you search in the map before, discard this comment

paul-nicolas (approved):

components/fctl/cmd/ledger/create.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	stack, err := fctl.ResolveStack(cmd, cfg, organizationID)
+	if err != nil {
+		return nil, err
+	}
+
+	if !fctl.CheckStackApprobation(cmd, stack, "You are about to create a new transaction") {
```
paul-nicolas, line 73: A new transaction ?

gfyrag: oups :D

paul-nicolas (approved):

PR #866, opened 2023-11-16
feat: allow options (listener or bindingAddress) with opts pattern

Dav-14:


libs/go-libs/httpserver/serverport.go
```diff
@@ -90,15 +90,50 @@ func StartServer(ctx context.Context, bind string, handler http.Handler, options
...
+
+func WithListener(listener net.Listener) serverOpts {
+	return func(server *server) {
+		server.listener = listener
+	}
+}
+
+func WithAddress(addr string) serverOpts {
+	return func(server *server) {
+		l, err := net.Listen("tcp", addr)
```
gfyrag, line 108: Put this inside StartServer function

Dav-14: In wich case ? In case the listener is not provided ?

PR #879, opened 2023-11-21
feat: Atlar connector and client

lwagner-getmomo:
# DESCRIPTION

This PR is currently a draft containing stubs.

The finished PR will contain an Atlar connector and an Atlar client (only as far as needed by the connector).

### DIFF
--- a/components/fctl/cmd/payments/connectors/install/atlar.go
+++ b/components/fctl/cmd/payments/connectors/install/atlar.go
@@ -1,0 +1,97 @@
+package install
+
+import (
+	"fmt"
+
+	"github.com/formancehq/fctl/cmd/payments/connectors/internal"
+	fctl "github.com/formancehq/fctl/pkg"
+	"github.com/formancehq/formance-sdk-go/pkg/models/operations"
+	"github.com/formancehq/formance-sdk-go/pkg/models/shared"
+	"github.com/pkg/errors"
+	"github.com/pterm/pterm"
+	"github.com/spf13/cobra"
+)
+
+type PaymentsConnectorsAtlarStore struct {
+	Success       bool   `json:"success"`
+	ConnectorName string `json:"connectorName"`
+}
+
+type PaymentsConnectorsAtlarController struct {
+	store              *PaymentsConnectorsAtlarStore
+	atlarAccessKeyFlag string
+	atlarSecretFlag    string
+}
+
+var _ fctl.Controller[*PaymentsConnectorsAtlarStore] = (*PaymentsConnectorsAtlarController)(nil)
+
+func NewDefaultPaymentsConnectorsAtlarStore() *PaymentsConnectorsAtlarStore {
+	return &PaymentsConnectorsAtlarStore{
+		Success: false,
+	}
+}
+
+func NewPaymentsConnectorsAtlarController() *PaymentsConnectorsAtlarController {
+	return &PaymentsConnectorsAtlarController{
+		store:              NewDefaultPaymentsConnectorsAtlarStore(),
+		atlarAccessKeyFlag: "access-key",
+		atlarSecretFlag:    "secret",
+	}
+}
+
+func NewAtlarCommand() *cobra.Command {
+	c := NewPaymentsConnectorsAtlarController()
+	return fctl.NewCommand(internal.AtlarConnector+" <access-key> <secret>",
+		fctl.WithShortDescription("Install an atlar connector"),
+		fctl.WithConfirmFlag(),
+		fctl.WithArgs(cobra.ExactArgs(2)),
+		fctl.WithStringFlag(c.atlarAccessKeyFlag, "", "Atlar API access key"),
+		fctl.WithStringFlag(c.atlarSecretFlag, "", "Atlar API secret"),
+		fctl.WithController[*PaymentsConnectorsAtlarStore](c),
+	)
+}
+
+func (c *PaymentsConnectorsAtlarController) GetStore() *PaymentsConnectorsAtlarStore {
+	return c.store
+}
+
+func (c *PaymentsConnectorsAtlarController) Run(cmd *cobra.Command, args []string) (fctl.Renderable, error) {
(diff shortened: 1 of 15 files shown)


components/payments/cmd/connectors/internal/connectors/atlar/Insomnium.json
lwagner-getmomo, line 1: I don't know if it's appropriate to put this here. I will gladly move this if necessary.

components/payments/cmd/connectors/internal/connectors/atlar/task_fetch_payments.go
```diff
@@ -0,0 +1,239 @@
...
+
+		paymentId := models.PaymentID{
+			PaymentReference: models.PaymentReference{
+				Reference: item.ID,
+				Type:      paymentType,
+			},
+			ConnectorID: connectorID,
+		}
+
+		batchElement := ingestion.PaymentBatchElement{
```
lwagner-getmomo, line 115: Since transactions can change over time, we'd want to overwrite any previously ingested payments. Currently this does not happen, instead trying to overwrite an ingested payment results in the following:

Screenshot 2023-11-28 at 16 54 24

@paul-nicolas do you have any idea on how to forcibly overwrite any payments already in the database?

In the future we might want to only fetch changes since the last poling. We'd have to save the time of last polling somewhere in order to make the connector resilient against interruptions of service. This is an optimization for later. Would we need to optimize this before merging, though?

lwagner-getmomo (commented):
Ingestation is largely done now, with the exception of the issue described in my comments and some refinements I laid out in comments in the code.

We also don't have examples for how vIBAN's behave right now, so I'd not expect the current version handling that case well.

components/payments/cmd/connectors/internal/connectors/atlar/DEVELOPERS.md
```diff
@@ -0,0 +1,9 @@
+# Atlar HTTP Client
+
+The directories `client` and `models` have been automatically generated by means of running
+
+```sh
+docker run --rm -it  --user $(id -u):$(id -g) -e GOPATH=$(go env GOPATH):/go -v $HOME:$HOME -w $(pwd) quay.io/goswagger/swagger generate client -f https://cdn.atlar.com/api/swagger.json -A client
```
paul-nicolas, line 6: No strong opinion here, but we're not fans of putting 250 files for the client part inside this repository, could it be possible to put them in another public repository while waiting for them to handle a go library on their sides ? 

What do you think ?

lwagner-getmomo: At least for `v1` Atlar is not going to provide a go library. Here's their response for reference:

Screenshot 2023-11-29 at 15 31 46

I'll ask Sonny about putting a standalone Atlar `v1` client up on a GH repo. We'll have to figure out the licensing. I can start looking into that.

components/fctl/cmd/payments/connectors/install/atlar.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	paymentsClient, err := fctl.NewStackClient(cmd, soc.Config, soc.Stack)
+	if err != nil {
+		return nil, err
+	}
+
+	response, err := paymentsClient.Payments.InstallConnector(cmd.Context(), operations.InstallConnectorRequest{
+		ConnectorConfig: shared.ConnectorConfig{
+			AtlarConfig: &shared.AtlarConfig{
```
paul-nicolas, line 71: AtlarConfig does not exist yet, you have to add it inside the openapi.yaml in the components/payments directory.

If you need any help on this matter, don't hesitate to ping me, I will be glad to help you !

lwagner-getmomo: Thanks for pointing that out! I assumed that would be generated. I'll add it and come back to you in case I run into problems.

I really appreciate you taking the time to take a look over this PR 🚀

components/payments/cmd/connectors/internal/connectors/atlar/task_fetch_payments.go
paul-nicolas: Actually, you have another error on balances from what I can see on your screenshot ! Let me check this error and go back to you. 

Otherwise, payments can be overwritten overtime (in order to update statuses etc...)

When ingesting payments, you can add a state in order be able to fetch only the last changes since the last polling. We are using it on Stripe actually.

components/fctl/cmd/payments/connectors/install/atlar.go
lwagner-getmomo: I added the missing definition of the connector config. Thanks again for pointing it out 🙌

components/payments/cmd/connectors/internal/connectors/atlar/DEVELOPERS.md
lwagner-getmomo: Atlar client is an external package, now.

components/payments/cmd/connectors/internal/connectors/atlar/task_fetch_payments.go
paul-nicolas: We fixed the balance problem, I think the next optimizations can be in other PRs to not have a big one at the end !

components/payments/cmd/connectors/internal/connectors/atlar/DEVELOPERS.md
paul-nicolas: Perfect thank you !

components/fctl/cmd/payments/connectors/install/atlar.go
paul-nicolas: Thank you !

paul-nicolas (approved):
Tested on my dev environment, everything works as expected. This PR just needs a rebase and then we can merge it and do the next parts/optimizations in other PRs !

Thank you !

PR #945, opened 2023-12-04
feat(payments): add payments creation endpoint

paul-nicolas:


components/payments/cmd/api/internal/api/payments.go
```diff
@@ -38,6 +39,61 @@ type paymentAdjustment struct {
...
+
+		if payment.SourceAccountID != nil {
+			data.SourceAccountID = payment.SourceAccountID.String()
+		}
+
+		if payment.DestinationAccountID != nil {
+			data.DestinationAccountID = payment.DestinationAccountID.String()
+		}
+
+		err = json.NewEncoder(w).Encode(api.BaseResponse[paymentResponse]{
```
gfyrag, line 87: maybe use shared lib

gfyrag (approved):

components/payments/cmd/api/internal/api/payments.go
paul-nicolas: Good idea, I will do a PR to unify all endpoints to use shared libs as they are all using the json encoder like that if you're okay with that !

PR #989, opened 2023-12-11
fix(payments): Correctly determine status of payment in Atlar connector

lwagner-getmomo:
# DESCRIPTION

I talked to the guys over at Atlar about what their transaction statuses mean (the documentation is hard to find). After some clarifications, here are some fixes to `determinePaymentStatus()`.

### DIFF
--- a/components/payments/cmd/connectors/internal/connectors/atlar/task_fetch_payments.go
+++ b/components/payments/cmd/connectors/internal/connectors/atlar/task_fetch_payments.go
@@ -169,10 +169,12 @@ func determinePaymentStatus(item *atlar_models.Transaction) models.PaymentStatus
 		// which was not yet reconciled with a payment from the statement
 		return models.PaymentStatusPending
 	}
-	if item.Reconciliation.Status == atlar_models.ReconciliationDetailsStatusRECONCILED {
+	if item.Reconciliation.Status == atlar_models.ReconciliationDetailsStatusBOOKED {
+		// A payment comissioned with the bank, which was not yet reconciled with a
+		// payment from the statement
 		return models.PaymentStatusPending
 	}
-	if item.Reconciliation.Status == atlar_models.ReconciliationDetailsStatusBOOKED {
+	if item.Reconciliation.Status == atlar_models.ReconciliationDetailsStatusRECONCILED {
 		return models.PaymentStatusSucceeded
 	}
 	return models.PaymentStatusOther


components/payments/cmd/connectors/internal/connectors/atlar/task_fetch_payments.go
```diff
@@ -169,12 +169,17 @@ func determinePaymentStatus(item *atlar_models.Transaction) models.PaymentStatus
...
+	if item.Reconciliation.Status == atlar_models.ReconciliationDetailsStatusBOOKED {
+		// A payment comissioned with the bank, which was not yet reconciled with a
+		// payment from the statement
 		return models.PaymentStatusPending
 	}
-	if item.Reconciliation.Status == atlar_models.ReconciliationDetailsStatusBOOKED {
+	if item.Reconciliation.Status == atlar_models.ReconciliationDetailsStatusRECONCILED {
 		return models.PaymentStatusSucceeded
 	}
+	if item.Reconciliation.Status == "RETURNED" { // atlar_models is missing an enum value
```
lwagner-getmomo, line 180: This value is also not documented, but they told me it does exist.

components/payments/cmd/connectors/internal/connectors/atlar/task_fetch_payments.go
lwagner-getmomo, line 1: I'm still investigating on cancelled payments.

components/payments/cmd/connectors/internal/connectors/atlar/task_fetch_payments.go
lwagner-getmomo: Apparently there was some confusion about payments and transactions. The mapping works as follows

- Formance `Payment` -> Atlar `Transaction`
- Formance `Payment Initiation` -> Atlar `Payment`

The payment object on one platform does not equal the payment object on the other platform. That led to some confusion. This PR still fixes something I just understood. I will fix updating `Payment Initiation` status in another PR.

components/payments/cmd/connectors/internal/connectors/atlar/task_fetch_payments.go
lwagner-getmomo: There was some confusion as described in https://github.com/formancehq/stack/pull/989#discussion_r1422479323.

Payments fetched via the transactions API cannot have a status of `Cancelled`. Only payment initiations can be cancelled. They will be subject of another PR.

paul-nicolas (approved):

PR #999, opened 2023-12-12
feat: add scopes in static clients for inner stack communication

Dav-14:


paul-nicolas (approved):
Let's just confirm the depends on that are optional as mentionned by @gfyrag

PR #1055, opened 2023-12-19
chore: Update Helm for regions

flemzord:


components/operator/config/default/kustomization.yaml
```diff
@@ -14,9 +14,9 @@ namePrefix: formance-system-
 buildMetadata: [originAnnotations]
 
 resources:
-- ../crd
+#- ../crd
+#- ../manager
```
paul-nicolas, line 18: Still in comments ?

helm/regions/values.yaml
```diff
@@ -22,10 +26,57 @@ agent:
...
     authentication:
         mode: bearer
         issuer: https://app.formance.cloud/api
         clientID: "REGION_ID"
-        clientSecret: ""
\ No newline at end of file
+        clientSecret: "REGION_SECRET"
+
+# External Services
+nats:
```
Dav-14, line 39: maybe let the user enabling them individually OR with the demo flag ?

components/operator/Earthfile
```diff
@@ -67,17 +67,14 @@ generate:
 helm-update:
     FROM core+builder-image
     RUN curl -s https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh | bash -s -- 4.5.4 /bin
-
     WORKDIR /src
     COPY --pass-args (+manifests/config) /src/config
     COPY --dir helm hack .
-
+    RUN rm -f helm/templates/crd/*
```
gfyrag, line 73: Why move this code BEFORE the kustomize build?
The file will not be removed at the end.

gfyrag (approved):

PR #1065, opened 2023-12-21
feat: Update Helm for regions & Add docs for CRD

flemzord:


components/operator/config.yaml
```diff
@@ -0,0 +1,17 @@
+processor:
```
gfyrag, line 1: can you rename this file?

gfyrag (approved):

PR #1101, opened 2024-01-08
feat: Add new Docker Image (scratch)

flemzord:


Earthfile
```diff
@@ -60,7 +60,7 @@ openapi:
     END
     FOR component IN $(cd ./ee && ls -d */)
         COPY (./ee/$component+openapi/src/ee/$component) /src/ee/$component
-    ENDle
```
gfyrag, line 63: :O

Earthfile
```diff
@@ -60,7 +60,7 @@ openapi:
...
+    END
     RUN toto
```
gfyrag, line 64: toto?

ee/gateway/scratch.Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM alpine:latest as certs
+RUN apk --update add ca-certificates
+
+FROM scratch
+COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+ADD https://raw.githubusercontent.com/formancehq/stack/main/ee/gateway/Caddyfile /etc/caddy/Caddyfile
```
gfyrag, line 6: Use main is dangerous!

gfyrag: Why not using file from filesystem?

PR #1132, opened 2024-01-17
fix(payments): fix atlar external ID when creating a payment

paul-nicolas:
## Description

When creating an Atlar payment, we want to serialize the external ID with the number of retries of the transfer, and not the length of the adjustements which changes if the payments is created on Atlar side, hence the error that the payment was not found when searching to update the status.


gfyrag (approved):

lwagner-getmomo:
I tested this and it worked as intended 🚀🎉

PR #1134, opened 2024-01-17
feat: add iam authentication for postgres

gfyrag:
Fixes ENG-491


components/ledger/libs/aws/iam/load.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"context"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+func InitFlags(flags *pflag.FlagSet) {
+	flags.String("aws-region", "", "Specify AWS region")
```
paul-nicolas, line 12: Only nitpicking, but should we put the flag name in constant ? (if we want to access them outside of the package one day ?)

components/ledger/libs/bun/bunconnect/flags.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/formancehq/stack/libs/go-libs/aws/iam"
+	"github.com/lib/pq"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+	"io"
+	"time"
+)
+
+func InitFlags(flags *pflag.FlagSet) {
+	flags.String("postgres-uri", "", "Postgres URI")
```
paul-nicolas, line 16: Same here

components/payments/cmd/api/flags.go
```diff
@@ -1 +0,0 @@
-package api
```
paul-nicolas, line 1: Only one line in this file ? Should we delete it ?

paul-nicolas (approved):

PR #1140, opened 2024-01-19
fix(ledger): pit usage

gfyrag:


components/ledger/internal/storage/ledgerstore/migrations/0-init-schema.sql
```diff
@@ -651,7 +651,7 @@ with all_assets as (select v.v as asset
                         join lateral (
                    select *
                    from moves s
-                   where (_before is null or s.insertion_date <= _before)
+                   where (_before is null or s.effective_date <= _before)
```
paul-nicolas, line 654: What about the old buckets ? Will the function still being fixed even with the migration already done before ?

paul-nicolas (approved):

PR #1146, opened 2024-01-23
fix: for future breaking change, and user properties addition

Dav-14:
Fixes ENG-518
- fix: for future breaking changes
- feat: rename command


components/fctl/cmd/stack/users/upsert.go
```diff
@@ -29,10 +29,11 @@ func NewUpsertController() *UpsertController {
 }
 
 func NewUpsertCommand() *cobra.Command {
```
gfyrag, line 31: Why `upsert` ?

components/fctl/cmd/stack/users/upsert.go
```diff
@@ -29,10 +29,11 @@ func NewUpsertController() *UpsertController {
...
-	return fctl.NewCommand("upsert <stack-id> <user-id> <stack-role>",
+	return fctl.NewCommand("upsert <stack-id> <user-id>",
 		fctl.WithAliases("usar"),
```
gfyrag, line 33: usar?

components/fctl/cmd/stack/users/upsert.go
Dav-14: Because the user could not exist in the organization.

Do we want to enforced the invitation process ?

components/fctl/cmd/cloud/organizations/users/link.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	role := fctl.GetString(cmd, "role")
+	req := membershipclient.UpdateOrganizationUserRequest{}
+	if role != "" {
+		req.Role = membershipclient.Role(role)
+	} else {
+		access, res, err := apiClient.DefaultApi.ReadUserOfOrganization(cmd.Context(), organizationID, args[0]).Execute()
```
paul-nicolas, line 90: I don't understand why you get the roles when the user doesn't specify any. You have two solutions right ?
- Either the user doesn't exists in the organization, and so this call will fail
- Or the user exists and you put the same role with the upsert.
I think it's better to return an error if no roles in specify in the commands, and just upsert if roles are specified, what do you think ?

paul-nicolas (approved):
Other than my two previous comments, i';m good with the PR

PR #1194, opened 2024-02-06
feat(fctl): update a stack

Dav-14:
- feat: add command update stack
- feat: add disclaimer


components/fctl/cmd/stack/update.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+}
+func NewStackUpdateController() *StackUpdateController {
+	return &StackUpdateController{
+		store: NewDefaultStackUpdateStore(),
+	}
+}
+
+func NewUpdateCommand() *cobra.Command {
+	return fctl.NewMembershipCommand("update <stack-id> [name]",
```
gfyrag, line 43: Why id and name?

components/fctl/cmd/stack/update.go
```diff
@@ -0,0 +1,195 @@
...
+		fctl.WithShortDescription("Update a created stack, name, or version"),
+		fctl.WithArgs(cobra.ExactArgs(2)),
```
gfyrag, line 45: update <stack-id> [name] indicate an optional arg

gfyrag (commented):
I guess we should add "fctl stacks upgrade" version, dedicated to upgrade the stack, and able to check the version (need to be greater than previous), and with status check (as when we create a stack).

components/fctl/cmd/stack/upgrade.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	apiClient, err := fctl.NewMembershipClient(cmd, cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(args) < 1 {
```
gfyrag, line 67: Cannot happened

components/fctl/cmd/stack/upgrade.go
```diff
@@ -0,0 +1,140 @@
...
+		return nil, errors.New("stack-id is required")
+	}
+
+	stack, res, err := apiClient.DefaultApi.GetStack(cmd.Context(), organization, args[0]).Execute()
+	if err != nil {
+		return nil, errors.Wrap(err, "retrieving stack")
+	}
+	if res.StatusCode > 300 {
```
gfyrag, line 75: It's wrong

components/fctl/cmd/stack/upgrade.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	req := membershipclient.StackVersion{
+		Version: nil,
+	}
+	specifiedVersion := fctl.GetString(cmd, versionFlag)
+	switch {
+	case httpResponse.StatusCode == http.StatusOK && specifiedVersion == "":
+		var options []string
+		for _, version := range availableVersions.Data {
```
gfyrag, line 91: Maybe filter higher versions?

Dav-14: I rather put the filters on membership side ?

gfyrag: Yes, it must be on membership side. But it can be on client side also. From a UI perspective, it is logic to filter values which will be refused by membership.

gfyrag (approved):

paul-nicolas (approved):

PR #1196, opened 2024-02-06
feat(payments): make Modulr connector stateful

ArnaudCalmettes:
fixes ENG-567

### DIFF
--- a/components/payments/cmd/connectors/internal/connectors/modulr/task_fetch_accounts.go
+++ b/components/payments/cmd/connectors/internal/connectors/modulr/task_fetch_accounts.go
@@ -44,7 +44,7 @@ func (s *fetchAccountsState) IsNew(account *client.Account) (bool, error) {
 
 func (s *fetchAccountsState) FilterNew(accounts []*client.Account) ([]*client.Account, error) {
 	// accounts are assumed to be sorted by creation date.
-	var firstNewIdx int
+	firstNewIdx := len(accounts)
 	for idx, account := range accounts {
 		isNew, err := s.IsNew(account)
 		if err != nil {
--- a/components/payments/cmd/connectors/internal/connectors/modulr/task_fetch_accounts.go
+++ b/components/payments/cmd/connectors/internal/connectors/modulr/task_fetch_accounts.go
@@ -20,16 +20,25 @@ import (
 )
 
 type fetchAccountsState struct {
-	LastCreatedDate string `json:"last_created_date"`
+	LastCreatedTime time.Time `json:"last_created_date"`
 }
 
-func (s *fetchAccountsState) SetLatest(latest *client.Account) error {
-	lastCreatedTime, err := time.Parse("2006-01-02T15:04:05.999-0700", latest.CreatedDate)
+func (s *fetchAccountsState) SetLatestAccount(latest *client.Account) error {
+	createdTime, err := time.Parse("2006-01-02T15:04:05.999-0700", latest.CreatedDate)
 	if err != nil {
 		return err
 	}
-	s.LastCreatedDate = lastCreatedTime.Format("2006-01-02T15:04:05-0700")
+	if createdTime.After(s.LastCreatedTime) {
+		s.LastCreatedTime = createdTime
+	}
 	return nil
+}
+
+func (s *fetchAccountsState) GetLastCreatedDate() string {
+	if s.LastCreatedTime.IsZero() {
+		return ""
+	}
+	return s.LastCreatedTime.Format("2006-01-02T15:04:05-0700")
 }
 
 func taskFetchAccounts(config Config, client *client.Client) task.Task {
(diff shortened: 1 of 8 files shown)


gfyrag (approved):

components/payments/cmd/connectors/internal/connectors/modulr/task_fetch_accounts.go
```diff
@@ -51,37 +73,42 @@ func fetchAccounts(
...
 			break
 		}
 
 		if err := ingestAccountsBatch(ctx, connectorID, ingester, pagedAccounts.Content); err != nil {
-			return err
+			return state, err
 		}
 
 		for _, account := range pagedAccounts.Content {
+			if err := state.SetLatest(account); err != nil {
```
paul-nicolas, line 93: Hoe do you make sure that accounts are sorted, so the the set latest always sets the last created at ?

components/payments/cmd/connectors/internal/connectors/modulr/task_fetch_accounts.go
```diff
@@ -51,37 +73,42 @@ func fetchAccounts(
...
+				return state, err
+			}
```
paul-nicolas, line 95: Also, usually when passing the last created at to the psp call, it will return all accounts with createat >= last created at, so you will process everytime the accounts with last created at

components/payments/cmd/connectors/internal/connectors/modulr/task_fetch_beneficiaries.go
```diff
@@ -15,13 +15,30 @@ import (
...
+	LastCreated string `json:"last_created"`
+}
+
+func (s *fetchBeneficiariesState) SetLatest(latest *client.Beneficiary) error {
+	lastCreatedTime, err := time.Parse("2006-01-02T15:04:05.999-0700", latest.Created)
+	if err != nil {
+		return err
+	}
+	creationDate := lastCreatedTime.Format("2006-01-02T15:04:05-0700")
+	if s.LastCreated == "" || creationDate > s.LastCreated {
```
paul-nicolas, line 28: I don't think comparing dates as strings here is a good idea, because of the timezone ("-0700") than can lead to same dates not being equal.

components/payments/cmd/connectors/internal/connectors/modulr/task_fetch_transactions.go
```diff
@@ -19,12 +19,29 @@ import (
...
+	LastTransactionDate string `json:"last_transaction_date"`
+}
+
+func (s *fetchTransactionsState) SetLatest(latest *client.Transaction) error {
+	lastTransactionTime, err := time.Parse("2006-01-02T15:04:05.999-0700", latest.TransactionDate)
+	if err != nil {
+		return err
+	}
+	tnDate := lastTransactionTime.Format("2006-01-02T15:04:05-0700")
+	if s.LastTransactionDate == "" || tnDate > s.LastTransactionDate {
```
paul-nicolas, line 32: Same comment here for the string comparaison

components/payments/cmd/connectors/internal/connectors/modulr/task_fetch_accounts.go
paul-nicolas: My bad I didn't see the sortFields and sortOrder

components/payments/cmd/connectors/internal/connectors/modulr/task_fetch_beneficiaries.go
ArnaudCalmettes: Fixed, they're now stored and compared as time.Time

components/payments/cmd/connectors/internal/connectors/modulr/task_fetch_accounts.go
ArnaudCalmettes: In that case, we might add another method to the state, in order to perform the comparison on our side and avoid processing the same thing twice.

components/payments/cmd/connectors/internal/connectors/modulr/task_fetch_transactions.go
ArnaudCalmettes: Fixed too

components/payments/cmd/connectors/internal/connectors/modulr/task_fetch_transactions.go
```diff
@@ -81,7 +167,7 @@ func fetchTransactions(
 		}
 	}
 
-	return nil
+	return state, nil
```
paul-nicolas, line 170: You should return newState here, when testing, the new state is never inserted in the database

components/payments/cmd/connectors/internal/connectors/modulr/task_fetch_accounts.go
```diff
@@ -19,13 +19,60 @@ import (
...
+
+func (s *fetchAccountsState) IsNew(account *client.Account) (bool, error) {
+	createdTime, err := time.Parse("2006-01-02T15:04:05.999-0700", account.CreatedDate)
+	if err != nil {
+		return false, err
+	}
+	return createdTime.After(s.LastCreatedTime), nil
+}
+
+func (s *fetchAccountsState) FilterNew(accounts []*client.Account) ([]*client.Account, error) {
```
paul-nicolas, line 45: This function is not working when there is no new element in the slice. The firstNewIdx will never be updated and will still be 0, so you will process again all accounts of the slice

ArnaudCalmettes: fixed. This index is now initialized with the slice's length.

paul-nicolas (approved):

PR #1210, opened 2024-02-08
fix: update ledger v1 docs about pageSize

igordreher:
Turns out the maximum page size is 1000
Also, the `/balances` endpoint accepts `pageSize`


gfyrag (approved):

gfyrag:
@igordreher Thks!

PR #1223, opened 2024-02-09
fix: stack crd readiness depend on namespace and dependent scheme

Dav-14:


components/operator/internal/core/controllers.go
```diff
@@ -15,7 +16,15 @@ type ObjectController[T client.Object] func(ctx Context, req T) error
 
 func ForObjectController[T v1beta1.Object](controller ObjectController[T]) ObjectController[T] {
 	return func(ctx Context, object T) error {
+
```
gfyrag, line 19: Remove logging please

components/operator/internal/core/controllers.go
```diff
@@ -15,7 +16,15 @@ type ObjectController[T client.Object] func(ctx Context, req T) error
...
+		logger, err := logr.FromContext(ctx)
```
paul-nicolas, line 20: Why introducing a new library for logging ?

gfyrag: Provided by the kubebuilder framework ;)

components/operator/internal/resources/ledgers/controller.go
```diff
@@ -42,7 +42,7 @@ import (
 //+kubebuilder:rbac:groups=batch,resources=cronjobs,verbs=get;list;watch;create;update;patch;delete
 
 func Reconcile(ctx Context, stack *v1beta1.Stack, ledger *v1beta1.Ledger, version string) error {
-
+	// return fmt.Errorf("not implemented")
```
gfyrag, line 45: remove this

components/operator/internal/resources/stacks/controller.go
```diff
@@ -38,6 +39,110 @@ import (
...
+	return NewApplicationError("No modules found")
+
+}
+
+func areDependentReady(ctx Context, stack *v1beta1.Stack) error {
+	pendingResources := make([]string, 0)
+	setInfo := map[string][]string{}
+	for _, rtype := range ctx.GetScheme().AllKnownTypes() {
+		v := reflect.New(rtype).Interface()
+		var r v1beta1.Dependent
```
gfyrag, line 76: this var is useless since the next line is `r, ok := ...`

components/operator/internal/resources/stacks/controller.go
```diff
@@ -46,6 +151,16 @@ func Reconcile(ctx Context, stack *v1beta1.Stack) error {
 		return err
 	}
 
+	err = hasSomeModule(ctx, stack)
```
gfyrag, line 154: This will return an error if the stack does not have any module.
From a business point of view, it's wrong. A stack with no modules is just ready.

gfyrag: You don't really need this, the `areDependentReady` is enough.

gfyrag: And why not?
A stack is just an objet, you can create a stack without any module, it is a fact.
And this code will leverage an error, where there are no errors.

gfyrag (approved):

PR #1228, opened 2024-02-12
feat: status can now be handle by membership on create, restore, upgrade

Dav-14:


components/fctl/cmd/stack/controller.go
```diff
@@ -7,27 +7,30 @@ import (
...
-		case <-time.After(time.Second):
+		if stack.Data.Status == "READY" {
+			return nil
 		}
+
+		if waitTime > time.Minute {
+			pterm.Warning.Println("Waiting for stack to be ready...")
+			return fmt.Errorf("there might a problem with the stack scheduling, retry and if the problem persists please contact the support")
+		}
+		time.Sleep(waitTime)
```
gfyrag, line 32: Avoid using Sleep, this will not allow to cancel the command

PR #1262, opened 2024-02-19
feat(operator): Add metrics for gateway

flemzord:
ENG-656


paul-nicolas (approved):

gfyrag (commented):
Maybe add a boolean to enable?
Some clients will not want the endpoint.

PR #1274, opened 2024-02-23
feat: wrong status ready check (wrong structure)

Dav-14:
- **feat: reduce wait time**
- **fix: wrong check**


components/fctl/cmd/stack/controller.go
```diff
@@ -16,7 +16,12 @@ func waitStackReady(cmd *cobra.Command, client *membershipclient.APIClient, prof
 	var err error
 	var stackRsp *membershipclient.CreateStackResponse
 
-	waitTime := 10 * time.Second
+	waitTime := 2 * time.Second
+	sum := 2 * time.Second
```
paul-nicolas, line 20: Sum is never used ?

paul-nicolas (approved):

PR #1275, opened 2024-02-23
feat(payments): handle precision from string

paul-nicolas:
Fixes ENG-707

### DIFF
--- a/components/payments/cmd/connectors/internal/connectors/currency/amount.go
+++ b/components/payments/cmd/connectors/internal/connectors/currency/amount.go
@@ -1,0 +1,120 @@
+package currency
+
+import (
+	"bytes"
+	"fmt"
+	"math/big"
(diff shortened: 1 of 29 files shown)


components/payments/cmd/connectors/internal/connectors/currency/amount.go
```diff
@@ -0,0 +1,120 @@
...
+	amountString := amount.String()
+	amountStringLength := len(amountString)
+
+	if precision == 0 {
+		// Nothing to do
+		return amountString, nil
+	}
+
+	decimalPart := bytes.NewBufferString("")
+	for p := precision; p > 0; p-- {
```
gfyrag, line 105: p >= 0 ?

gfyrag (approved):
Seems complicated but if it works

PR #1285, opened 2024-02-26
fix(orchestration): stop trigger workflows when variables evaluation fails

gfyrag:
Fixes ENG-713


ee/orchestration/internal/storage/migrations.go
```diff
@@ -116,6 +116,17 @@ add primary key (instance_id, stage, temporal_run_id);
 			return nil
 		},
 	},
+	{
```
gfyrag, line 119: Add a new column 'error' on table "triggers_occurrences"

ee/orchestration/internal/triggers/activity.go
```diff
@@ -74,35 +76,42 @@ func (a Activities) ProcessTrigger(ctx context.Context, trigger Trigger, request
 
 	span := trace.SpanFromContext(ctx)
 	var (
-		evaluated map[string]string
-		err       error
+		evaluated    map[string]string
```
gfyrag, line 79: There, we return a specific error on variables evaluation failures. This error is non retryable.

ee/orchestration/openapi/v2.yaml
```diff
@@ -659,8 +659,6 @@ components:
       type: object
       required:
         - triggerID
-        - workflowInstanceID
```
gfyrag, line 662: Imply a breaking changes on the sdk. A simple value will become a pointer? Is this ok with the actual version of orchestration? @flemzord

paul-nicolas: Is FCTL still compiling after this change ?

ee/orchestration/internal/triggers/workflow_trigger.go
```diff
@@ -42,7 +45,10 @@ func (w triggerWorkflow) RunTrigger(ctx temporalworkflow.Context, req ProcessEve
 			req,
 		).Get(ctx, nil)
 		if err != nil {
-			return err
+			applicationError := &temporal.ApplicationError{}
+			if !errors.As(err, &applicationError) {
```
paul-nicolas, line 49: What about the other type of errors ? We will return nil, is that okay ?

paul-nicolas (approved):

paul-nicolas (approved):

PR #1294, opened 2024-02-28
feat(gateway): expose events

gfyrag:
Fixes ENG-709

### DIFF
--- a/components/operator/go.mod
+++ b/components/operator/go.mod
@@ -1,11 +1,12 @@
 module github.com/formancehq/operator
 
-go 1.21
+go 1.21.6
 
-toolchain go1.21.5
+toolchain go1.22.0
 
 require (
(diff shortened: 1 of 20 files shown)


components/operator/go.mod
```diff
@@ -1,11 +1,12 @@
 module github.com/formancehq/operator
 
-go 1.21
+go 1.21.6
 
-toolchain go1.21.5
+toolchain go1.22.0
```
flemzord, line 5: Go version is upgraded ?

go.mod
```diff
@@ -1,3 +1,3 @@
 module github.com/formancehq/stack
 
-go 1.20
+go 1.22
```
flemzord, line 3: Go upgrade ?

components/operator/go.mod
gfyrag: I'll change

PR #1296, opened 2024-02-29
feat: Opensearch with AWS IAM

Dav-14:
search, benthos, and operator configuration,


components/operator/internal/resources/benthos/controller.go
```diff
@@ -93,8 +93,14 @@ func createDeployment(ctx Context, stack *v1beta1.Stack, b *v1beta1.Benthos) err
...
 
+	serviceAccountName, err := settings.GetAWSServiceAccount(ctx, stack.Name)
+	if err != nil {
+		return err
+	}
+
+	awsIAMEnabled := serviceAccountName != ""
 	var resourceReference *v1beta1.ResourceReference
-	if secret := elasticSearchURI.Query().Get("secret"); secret != "" {
+	if secret := elasticSearchURI.Query().Get("secret"); !awsIAMEnabled && secret != "" {
```
paul-nicolas, line 103: So you're not creating any secret if iam is enabled, am I right ?

ee/search/benthos/resources/output_elasticsearch.yaml
```diff
@@ -1,42 +1,45 @@
...
+        password: ${BASIC_AUTH_PASSWORD}
+      batching:
+        count: ${OPENSEARCH_BATCHING_COUNT:10}
+        period: ${OPENSEARCH_BATCHING_PERIOD:1s}
+      multipart: true
+      doc: ${! if json().exists("document") { json("document") } else {""} }
+      script: ${! if json().exists("script") { json("script") } else {""} }
+      upsert: ${! if json().exists("upsert") { json("upsert") } else {""} }
+      params: ${! if json().exists("params") { json("params") } else {""} }
+      aws:
```
paul-nicolas, line 43: ```It's possible to enable AWS connectivity with this output using the aws fields. However, you may need to set sniff and healthcheck to false for connections to succeed.```

This was written in the opensearch doc, should we check that it's working without setting to false the healthcheck ?

ee/search/benthos/resources/output_elasticsearch.yaml
```diff
@@ -1,42 +1,45 @@
...
+        enabled: ${AWS_IAM_ENABLED:false}
+        region: ${AWS_REGION:""}
```
paul-nicolas, line 45: Should we also set the `aws.credentials` struct here ? or will it fetch automatically the right env var ?

components/operator/internal/resources/benthos/controller.go
Dav-14: Yes !

paul-nicolas (approved):

PR #1300, opened 2024-03-01
feat(payments): add system to retry task or not

paul-nicolas:
Fixes ENG-762

### DIFF
--- a/components/payments/cmd/connectors/internal/task/scheduler_test.go
+++ b/components/payments/cmd/connectors/internal/task/scheduler_test.go
@@ -2,7 +2,6 @@ package task
 
 import (
 	"context"
-	"fmt"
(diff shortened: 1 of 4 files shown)


components/payments/cmd/connectors/internal/task/scheduler_test.go
```diff
@@ -36,6 +37,8 @@ func TaskTerminatedWithStatus(
 			return false
 		}
 
+		fmt.Println(status, resultErr, errString, expectedStatus)
```
Dav-14, line 40: Println ?
In test OK

Dav-14 (approved):
Seems OK for me, i have not the all the scope. I've saw much panic(err) in startTask

PR #1305, opened 2024-03-04
feat(agent): add additional label on existing stack order

Dav-14:
- **feat: add additional labels dict on sync existing stacks**
- **feat(agent): test adding additional label on existing stack order**


ee/agent/server.proto
```diff
@@ -54,6 +54,7 @@ message Stack {
   bool disabled = 6;
   google.protobuf.Timestamp deletedAt = 7;
   string versions = 8;
+  map<string, string> additionalLabels = 10;
```
Dav-14, line 57: Add the label on the 10th slot,
I've reserved the 9 for an other feature

ee/agent/internal/membership_listener.go
```diff
@@ -284,7 +289,7 @@ func (c *membershipListener) enableStack(ctx context.Context, stack *generated.E
 	sharedlogging.FromContext(ctx).Infof("Stack %s enabled", stack.ClusterName)
 }
 
-func (c *membershipListener) createOrUpdate(ctx context.Context, gvk schema.GroupVersionKind, name string, owner *metav1.OwnerReference, content map[string]any) (*unstructured.Unstructured, error) {
+func (c *membershipListener) createOrUpdate(ctx context.Context, gvk schema.GroupVersionKind, name string, owner *metav1.OwnerReference, content map[string]any, additionalLabel map[string]any) (*unstructured.Unstructured, error) {
```
paul-nicolas, line 292: why passing map[string]any since it's a map[string]string ?

paul-nicolas (approved):

ee/agent/internal/membership_listener.go
paul-nicolas: Alright it's kube that want that !

PR #1318, opened 2024-03-12
fix(payments): no states for accounts and balances for modulr

paul-nicolas:
Fixes ENG-815


Dav-14 (commented):
Why remove fetchAccountsState ? Is it related to something else ?

Dav-14 (approved):

PR #1319, opened 2024-03-12
fix(payments): fix missing provider in account responses

paul-nicolas:
Fixes ENG-812


releases/sdks/go/gen.yaml
```diff
@@ -7,7 +7,7 @@ generation:
     nameResolutionDec2023: false
   telemetryEnabled: false
 go:
-  version: 0.0.1
+  version: v0.0.0
```
Dav-14, line 10: speakeasy ?

Dav-14 (approved):

releases/sdks/go/gen.yaml
paul-nicolas: Oh it was overrided by the speakeasy command, let me put that back

PR #1321, opened 2024-03-13
fix(orchestration): always init search attributes

paul-nicolas:


paul-nicolas:
We launched the temporal CLI to fix the typo directly

PR #1335, opened 2024-03-18
bug(wallets): properly handle credit/debit inputs to avoid status code 500

gfyrag:
Fix ENG-858


components/ledger/internal/account_test.go
```diff
@@ -86,7 +88,7 @@ func TestValidateAddress(t *testing.T) {
 	for _, testCase := range testsCases {
 		testCase := testCase
 		t.Run(testCase.name, func(t *testing.T) {
-			require.Equal(t, testCase.shouldBeOk, ValidateAddress(testCase.address))
+			require.Equal(t, testCase.shouldBeOk, accounts.ValidateAddress(testCase.address))
```
paul-nicolas, line 91: Should this test go to pkg/account_test instead of staying here ?

paul-nicolas (approved):

PR #1354, opened 2024-03-25
feat(operator): manage namespace if not exist

Dav-14:
## Summary by CodeRabbit

- **New Features**
	- Introduced a configuration option to control automatic namespace creation for stacks.
- **Refactor**
	- Improved internal logic for handling namespace creation and cleanup based on new configuration settings.
- **Documentation**
	- Added a new constant for labeling created resources.

### DIFF
--- a/components/operator/internal/resources/stacks/controller.go
+++ b/components/operator/internal/resources/stacks/controller.go
@@ -2,12 +2,14 @@ package stacks
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"sort"
 	"strings"
 
 	"github.com/formancehq/operator/api/formance.com/v1beta1"
+	"github.com/formancehq/operator/internal/core"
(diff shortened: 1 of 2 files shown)


gfyrag (changes requested):
Too complicated.
Just create the ns if not present.
Otherwise, don't do anything.

Dav-14:
I know use Controller Reference which delete the namespace if added

If already exist -> Do nothing
If created by agent -> we create the namespace with the ref

components/operator/internal/resources/stacks/controller.go
```diff
@@ -188,13 +203,6 @@ func Clean(ctx Context, t *v1beta1.Stack) error {
 	}
 
 	logger.Info("All dependencies removed, remove namespace")
```
gfyrag, line 205: Change log

gfyrag (approved):

PR #1403, opened 2024-04-09
feat: decrease batch size when writing logs

gfyrag:


Dav-14 (commented):
We only can have 50 insertion in the batcher ?
Is it by ledger ? Bucket ?

paul-nicolas (approved):

PR #1439, opened 2024-04-23
feat(flows): ENG-878-Ability to name triggers in workflows

agourdel:
fixes ENG-878


gfyrag (commented):
Please rebase, there are some notification on payments in this PR

ee/orchestration/openapi.yaml
```diff
@@ -2099,6 +2125,8 @@ components:
         vars:
           type: object
           additionalProperties: true
+        name:
+          type: string
```
Dav-14, line 2129: Maybe make it required as the default is ""

PR #1444, opened 2024-04-24
fix(fctl): ENG-859 revert transaction returns empty error

agourdel:


agourdel:
Rebase Has been done.

PR #1449, opened 2024-04-25
feat(ledger): ENG-986 ability to filter on a metadata existence

agourdel:
Add filter $exists: { metadata: XXX} to endpoints accounts, transactions, volumes and balances


gfyrag (approved):

agourdel:
@gfyrag "Main" check failed because secret are not injected in Workflow when PR comes  from outside the organization.

PR #1520, opened 2024-05-21
fix(operator): skipped condition removed too early

gfyrag:


Dav-14 (changes requested):
stack.Ready is always ready in structure definition

PR #1525, opened 2024-05-22
feat(operator): Add Payments Connector Statefull & Replicas use for ledger

flemzord:


components/operator/internal/resources/ledgers/deployments.go
```diff
@@ -27,17 +27,34 @@ import (
 func installLedger(ctx core.Context, stack *v1beta1.Stack,
 	ledger *v1beta1.Ledger, database *v1beta1.Database, image string, isV2 bool) error {
 
+	replicasSettings, err := settings.GetStringOrDefault(ctx, stack.Name, "1", "deployments", "ledger", "replicas")
```
gfyrag, line 30: Why not using GetIntOrDefault ?
More accurate for this value I think

components/operator/internal/resources/ledgers/deployments.go
```diff
@@ -27,17 +27,34 @@ import (
...
 			return err
 		}
 		return installLedgerMonoWriterMultipleReader(ctx, stack, ledger, database, image, isV2)
-	default:
+	case v1beta1.DeploymentStrategySingle:
 		if err := uninstallLedgerMonoWriterMultipleReader(ctx, stack); err != nil {
 			return err
 		}
 		return installLedgerSingleInstance(ctx, stack, ledger, database, image, isV2)
+	default:
```
gfyrag, line 46: Ugly copy/paste :D

You could have define a variable `isSingle` before the switch to determine the mode following either the crd property or the setting.
Then replace the switch by a if/else.

components/operator/internal/resources/payments/deployments.go
```diff
@@ -253,6 +253,7 @@ func createConnectorsDeployment(ctx core.Context, stack *v1beta1.Stack, payments
 				},
 			},
 		}).
+		WithStateful(true).
```
gfyrag, line 256: Bien vu!

gfyrag (approved):

PR #1531, opened 2024-05-23
fix(fctl): unwrap loop to check if there is sdk errors

Dav-14:


gfyrag (changes requested):
You don't check the first level of error

gfyrag (approved):

paul-nicolas (approved):

PR #1568, opened 2024-06-20
feat: add parallel bulk processing

altitude:


components/ledger/internal/api/v2/bulk.go
```diff
@@ -42,168 +41,193 @@ type Result struct {
...
 			ErrorCode:        code,
 			ErrorDescription: err.Error(),
 			ResponseType:     "ERROR",
-		})
+		}
 		errorsInBulk = true
 	}
 
+	var bulkSuccess = func(index int, action string, data any) {
+		ret[index] = Result{
```
gfyrag, line 197: Since you allow parallelism, you need to use a lock on the `ret` variable

PR #1591, opened 2024-07-02
feat(wallets): add ik support

gfyrag:
Fixes ENG-1122


ee/wallets/pkg/ledger_interface.go
```diff
@@ -193,7 +193,8 @@ func (d DefaultLedger) CreateTransaction(ctx context.Context, ledger string, tra
 				return &transaction.Timestamp.Time
 			}(),
 		},
-		Ledger: ledger,
+		Ledger:         ledger,
+		IdempotencyKey: pointer.For(ik),
```
paul-nicolas, line 197: What if the ik is "" ? Should we not add anything in this case ?

gfyrag: The ledger will handle the empty string in the same way

paul-nicolas (approved):

paul-nicolas (approved):

PR #1595, opened 2024-07-04
feat(webhooks): Refonte V2 Webhooks

agourdel:
ENG-989


ee/webhooks/internal/commons/attempt.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func SetSuccesStatus(a *Attempt) {
+	a.Status = SuccessStatus
+	a.DateStatus = time.Now()
+}
+
+func SetNextRetry(a *Attempt) {
+	now := time.Now()
+	a.NextTry = now.Add(time.Duration(1<<a.NbTry) * time.Second)
```
gfyrag, line 111: Maybe set a max ? After some try, the delay will be very long.

gfyrag: Also, maybe, we should allow to not have a max number of attempts.

agourdel: Effectivement mais deux écoles :
On peut dire que comme le nombre de tentative max est géré par l'application, (par un paramètre) forcément le temps d'attente maximum entre deux attempts sera borné.
Ou rajouter ici un max delay. 

MAIS, je sens venir gros comme  une maison que le max delay choisi sera jamais le bon....

ee/webhooks/cmd/fx-modules/fx-modules.go
```diff
@@ -0,0 +1,25 @@
...
+
+	"go.opentelemetry.io/otel"
+)
+
+var Tracer = otel.Tracer("webhook")
+
+func FxProvideHttpClient() *http.Client {
+	return &http.Client{
+		Transport: otelhttp.NewTransport(http.DefaultTransport, otelhttp.WithSpanNameFormatter(func(operation string, r *http.Request) string {
+			str := fmt.Sprintf("%s %s", r.Method, r.URL.Path)
```
gfyrag, line 17: Avoid this if you have some kind of id in paths, like uuid. (I have not checked)
Span names are used as bucket.
If it changes always, the tooling cannot do the right job.
Sometimes, for some tools, it has a big performance impact.

ee/webhooks/cmd/fx-modules/fx-modules.go
```diff
@@ -0,0 +1,25 @@
...
+			if len(r.URL.Query()) == 0 {
+				return str
+			}
+			return fmt.Sprintf("%s?%s", str, r.URL.Query().Encode())
```
gfyrag, line 21: Same for query params. Worse then :D
You should use resource attributes to store those kind of data.

ee/webhooks/cmd/webhook-all-in-one.go
```diff
@@ -0,0 +1,186 @@
...
+				serverParams := httpserver.DefaultServerParams{}
+				serverParams.Addr = viper.GetString(flag.Listen)
+				serverParams.Auth = auth
+				serverParams.Info = serviceInfo
+				serverParams.Logger = logger
+
+				return serverParams
+			},
+
+			func(db *bun.DB) *storage.PostgresStore {
```
gfyrag, line 89: In Go, you can use directly some things like `fx.Provide(storage.NewPostgresStoreProvider)` instead of wrapping the call.

ee/webhooks/cmd/fx-modules/fx-modules.go
agourdel: This is code coming from V1.
I just used it as if because I Thought it was legacy.

libs/go-libs/sync/queue.go
```diff
@@ -0,0 +1,26 @@
+package sync
+
+type noCopy struct{}
+
+func (*noCopy) Lock()   {}
+func (*noCopy) Unlock() {}
+
+type Queue struct {
+	noCopy noCopy
```
gfyrag, line 9: Why are you doing that? I'm curious

libs/go-libs/sync/shared/sharedarr.go
```diff
@@ -0,0 +1,228 @@
...
+
+func (s *SharedArr[T]) UnsafeRemove(new *Shared[T]) *Shared[T] {
+	idx := s.UnsafeFind(new)
+
+	if idx == -1 {
+		return nil
+	}
+
+	if idx == 0 {
+		*s.Val = (*s.Val)[1:]
```
gfyrag, line 39: I think this will fail if the array has a size of 1 (I have a doubt)

libs/go-libs/sync/queue.go
agourdel: Do you mean the Queue struct ? 
Or the noCopy ?

libs/go-libs/sync/shared/sharedmaparr.go
```diff
@@ -0,0 +1,61 @@
+package shared
+
+type SharedMapArr[T any] struct {
+	Shared[map[string]*SharedArr[T]]
```
gfyrag, line 4: I'm not against generalizing some kind of utility around object sharing, but this start to sound like very specific to a specific use case ^^

ee/webhooks/cmd/fx-modules/fx-modules.go
Dav-14: As i remember, i've done the same on membership, r.URL.Path return the templated string "/test/{id}". It can be easily checked with Jeager

tests/integration/suite/webhooks-ledger-committed-transaction-worker.go
```diff
@@ -50,10 +53,13 @@ var _ = WithModules([]*Module{modules.Ledger, modules.Webhooks}, func() {
 		)
 		Expect(err).ToNot(HaveOccurred())
 		Expect(response.StatusCode).To(Equal(http.StatusOK))
+		
 	})
 
 	When("creating a transaction", func() {
+		
 		BeforeEach(func() {
+			time.Sleep(1*time.Second)
```
Dav-14, line 62: We should avoid this in tests. Tests should not depend on time. 

E.G You could implement a Ticker with manual declencher

ee/webhooks/internal/components/webhook_controller/controllers/utils/utils.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+}
+
+const (
+	ErrValidation  = "VALIDATION_TYPE"
+	ErrHealthcheck = "HEALTHCHECK_STATUS"
+)
+
+func DecodeJSONBody(r *http.Request, v any) error {
+	data, err := io.ReadAll(r.Body)
```
paul-nicolas, line 69: Quick opti, you can use json.NewDecoder(r.Body).Decode, instead of io.readall and unmarshall. It has its own buffering and for big body, it's actually better in performances than to raed everything at once in memory

paul-nicolas:
Should we also add some code in the operator to deploy webhooks ?

ee/webhooks/cmd/fx-modules/fx-modules.go
agourdel: resolve

ee/webhooks/cmd/fx-modules/fx-modules.go
agourdel: resolve

libs/go-libs/sync/shared/sharedarr.go
agourdel: In fact it works with Size == 1 but it failed with Size == 0
But Size == 0 is not a possible case here.

tests/integration/suite/webhooks-ledger-committed-transaction-worker.go
agourdel: This is orchestration tests, I need let the caches of the worker and collector runner refresh.
Whether with a sleep() or a ticker, I have to wait at least one second to be able to trigger the rest.
(And ticker is wrapper for timer by the way, with a sleep() inside)

So I think I do not understand how to do it without wait.

libs/go-libs/bun/bunmigrate/command.go
```diff
@@ -4,9 +4,7 @@ import (
 	"github.com/formancehq/stack/libs/go-libs/bun/bunconnect"
 	"github.com/spf13/cobra"
 	"github.com/uptrace/bun"
-
 	// Import the postgres driver.
-	_ "github.com/lib/pq"
```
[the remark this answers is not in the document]
agourdel: Well, All the integrations tests (for each modules) are working. 
I don't remember why I had to delete this line, because it goes back a long way  (I discussed with Gfyrag about it)
But I think I was able to delete it because it's not used.

ee/webhooks/internal/tests/components/webhook_controller/v2_controllers_hook_test.go
```diff
@@ -0,0 +1,228 @@
+package webhookcontroller
```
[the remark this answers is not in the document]
agourdel: Every test files are now in the same package than the code which is tested.

ee/webhooks/internal/tests/utils.go
```diff
@@ -0,0 +1,151 @@
...
+		return provider, err
+	}
+
+	return provider, nil
+
+}
+
+func fillDBWithData(db storage.PostgresStore, nbHooks int, nbAttempts int) error {
+
+	// hooks := make([]*commons.Hook, 0)
```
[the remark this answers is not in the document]
agourdel: Have been deleted.

ee/webhooks/internal/services/storage/postgres/postgres.go
```diff
@@ -0,0 +1,98 @@
...
+					if err == sql.ErrConnDone {
+						continue
+
+					}
+					//TODO(CriticPolitic)
+				}
+				for _, log := range logs {
+					event, err := commons.Event{}.FromPayload(log.Payload)
+					if err != nil {
+						//TODO(LOG) QUOI FAIRE ?
```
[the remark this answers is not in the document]
agourdel: Done

ee/webhooks/internal/components/webhook_controller/controllers/utils/utils.go
agourdel: Done

ee/webhooks/internal/components/webhook_controller/controllers/v2/controllers-attempts.go
```diff
@@ -0,0 +1,283 @@
...
+	storeInterface "github.com/formancehq/webhooks/internal/services/storage/interfaces"
+)
+
+const (
+	attemptPageSize int = 64
+)
+
+func RegisterV2AttemptControllers(serverHttp serverInterfaces.IHTTPServer, database storeInterface.IStoreProvider) {
+
+	serverHttp.Register(string(r.V2GetWaitingAttempts.Method), r.V2GetWaitingAttempts.Url, func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
agourdel: A Router/Handler (based on mux) logic has been created now exactly like other components of the stack.

ee/webhooks/openapi.yaml
```diff
@@ -224,14 +211,453 @@ paths:
...
+      tags:
+        - Webhooks
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/V2HookBodyParams'
+        required: true
+      responses:
+        "200":
```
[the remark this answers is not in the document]
agourdel: Need to be done. (Tomorrow)
(For every creation request)

ee/webhooks/internal/services/httpserver/default_server.go
```diff
@@ -0,0 +1,106 @@
...
+			handler.ServeHTTP(w, r)
+		})
+	})
+	defaultServer.handler.Use(func(handler http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			handler.ServeHTTP(w, r.WithContext(logging.ContextWithLogger(r.Context(), logger)))
+		})
+	})
+	defaultServer.handler.Group(func(r chi.Router) {
+		r.Get(HealthcheckRoute, func(_ http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
agourdel: fixed

ee/webhooks/internal/components/webhook_controller/controllers/v1/controllers-hook.go
```diff
@@ -0,0 +1,392 @@
+package controllers
```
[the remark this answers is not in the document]
agourdel: Done

PR #1607, opened 2024-07-15
Add more bank details to Accounts (internal) and Payments

lwagner-getmomo:
We're currently preparing for adding a second bank as  partner to work with.

Going forward, we will need to differentiate payments and internal accounts by bank. This PR introduces the metadata necessary to do so.

### DIFF
--- a/components/payments/cmd/connectors/internal/connectors/atlar/currencies.go
+++ b/components/payments/cmd/connectors/internal/connectors/atlar/currencies.go
@@ -5,5 +5,6 @@ import "github.com/formancehq/payments/cmd/connectors/internal/connectors/curren
 var (
 	supportedCurrenciesWithDecimal = map[string]int{
 		"EUR": currency.ISO4217Currencies["EUR"], //  Euro
+		"DKK": currency.ISO4217Currencies["DKK"],
 	}
 )
--- a/components/payments/cmd/connectors/internal/connectors/atlar/task_fetch_transactions.go
+++ b/components/payments/cmd/connectors/internal/connectors/atlar/task_fetch_transactions.go
@@ -49,7 +49,7 @@ func FetchTransactionsTask(config Config, client *client.Client) task.Task {
 
 			token = pagedTransactions.Payload.NextToken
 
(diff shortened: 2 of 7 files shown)


components/payments/cmd/connectors/internal/connectors/atlar/currencies.go
```diff
@@ -5,5 +5,6 @@ import "github.com/formancehq/payments/cmd/connectors/internal/connectors/curren
 var (
 	supportedCurrenciesWithDecimal = map[string]int{
 		"EUR": currency.ISO4217Currencies["EUR"], //  Euro
+		"DKK": currency.ISO4217Currencies["DKK"],
```
lwagner-getmomo, line 8: We need the `DKK` currency, because there is only one bank for the `EUR` currency in the Atlar default sandbox. After asking them for a second `EUR` bank, they asked us if we could use a `DKK` accounts for testing multi banking.

components/payments/cmd/connectors/internal/connectors/atlar/task_fetch_transactions.go
```diff
@@ -91,9 +93,20 @@ func ingestPaymentsBatch(
...
 	transaction *atlar_models.Transaction,
+	client *client.Client,
 ) (*ingestion.PaymentBatchElement, error) {
+	ctx, span := connectors.StartSpan(
+		ctx,
+		"atlar.atlarTransactionToPaymentBatchElement",
+		attribute.String("connectorID", connectorID.String()),
+		attribute.String("taskID", taskID.String()),
+	)
+	defer span.End()
```
lwagner-getmomo, line 108: I hope creating instead of passing a span is the correct thing to do here.

paul-nicolas (approved):

components/payments/cmd/connectors/internal/connectors/atlar/task_fetch_transactions.go
paul-nicolas: Yep it's good, it's gonna be linked to the parent span thanks to the StartSpan function, so LGTM

PR #1614, opened 2024-07-19
fix(ledger): add unique index on ik

gfyrag:


components/ledger/internal/storage/ledgerstore/migrations/7-add-ik-unique-index.sql
```diff
@@ -0,0 +1,17 @@
+update logs
+set idempotency_key = null
+where idempotency_key = '';
+
+update logs
+set idempotency_key = null
```
paul-nicolas, line 6: Here, we drop all ik with count > 1. Should we not keep one ? Otherwise, we will loose some ik

gfyrag: Hum, yes we could.
I don't know to write the query.

paul-nicolas (approved):

paul-nicolas (approved):

PR #1616, opened 2024-07-22
fix(ledger): Decode parameters in the ledger

flemzord:


Dav-14 (approved):

gfyrag (commented):
Use 404 instead of 500 on url parsing error (or any 4XX i don't care).

gfyrag (changes requested):
Use 404 instead of 500 on url parsing error (or any 4XX i don't care).

gfyrag (approved):
WithDetails is redundant. The error message will be duplicated in the response.

PR #1620, opened 2024-07-26
feat(sdk): Rework for use version in namespace

flemzord:


ee/auth/openapi.yaml
```diff
@@ -1,6 +1,6 @@
 openapi: 3.0.3
 info:
-  title: Auth API
```
gfyrag, line 3: find and replace?

gfyrag (approved):

PR #1622, opened 2024-07-29
fix(reconciliation): fix missing payments balances

paul-nicolas:
Assets missing and assets at 0 should be considered equivalent when doing reconciliation.

Fixes ENG-1175


ee/reconciliation/internal/api/service/reconciliation.go
```diff
@@ -163,6 +165,25 @@ func (s *Service) computeDrift(
 	return nil
 }
 
+// Missing asset should be considered as asset with balance 0
+func harmonizeBalances(ledgerBalances, paymentsBalances map[string]*big.Int) (map[string]*big.Int, map[string]*big.Int) {
```
altitude, line 169: 👍

altitude (approved):

PR #1627, opened 2024-07-31
fix(fctl): create organization with/without default role

Dav-14:


components/fctl/cmd/cloud/organizations/create.go
```diff
@@ -52,20 +53,27 @@ func (c *CreateController) Run(cmd *cobra.Command, args []string) (fctl.Renderab
...
+	}
+
+	if defaultStackRole != "" {
+		orgData.DefaultStackAccess = membershipclient.Role(defaultStackRole).Ptr()
+	}
+
+	if domain != "" {
+		orgData.Domain = pointer.For(domain)
+	}
+
```
Dav-14, line 73: those are optional

paul-nicolas (approved):

PR #1635, opened 2024-08-05
feat(ledger): detach context when creating transaction

paul-nicolas:


components/ledger/internal/api/v2/controllers_transactions.go
```diff
@@ -101,7 +102,10 @@ func postTransaction(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	res, err := l.CreateTransaction(r.Context(), getCommandParameters(r), *payload.ToRunScript())
+	ctx, cancel := contextutil.Detached(r.Context())
+	defer cancel()
```
gfyrag, line 106: Maybe never cancel this context?

paul-nicolas: Done

flemzord (approved):

PR #1638, opened 2024-08-06
fix(ledger): use a unique index on idempotency key and ledger

paul-nicolas:


tests/integration/suite/ledger-create-transaction-ik.go
```diff
@@ -0,0 +1,86 @@
...
+		})
+		Expect(err).To(BeNil())
+		Expect(createLedgerResponse.StatusCode).To(Equal(http.StatusNoContent))
+	})
+	When("creating a transaction on a ledger", func() {
+		var (
+			timestamp  = time.Now().Add(-1 * time.Minute).Round(time.Second).UTC()
+			timestamp2 = time.Now().Round(time.Second).UTC()
+		)
+		It("should fail", func() {
```
gfyrag, line 37: should fail?

gfyrag (approved):

PR #1641, opened 2024-08-06
feat(tools): Lock only active stacks

flemzord:


gfyrag (changes requested):
The operator do some operations even on disabled stack.

PR #1649, opened 2024-08-21
fix(operator): image translate 

Dav-14:


paul-nicolas (approved):
LGTM but maybe check with @gfyrag If I missed something !

gfyrag (approved):

PR #1665, opened 2024-08-29
feat: avoid tracking unbounded balances

ascandone:
Linear ticket: https://linear.app/formance/issue/ENG-1303/avoid-tracking-accounts-balances-when-not-necessary
TL;DR: we are trying to avoid locking balances for unbounded sources, in order to improve perf.

### Locking strategy before the PR:
readlocks were requested for each account that appeared in the program (excluding @world)
writelocks were requested for each account that appeared in the sources (excluding @world)

### Locking strategy after the PR: 
writelocks are requested for each _bounded_ account (that is: accounts that aren't world or aren't unbounded overdrafts) that appears in sources.
readlocks are requested for each account that appears in destinations or in the `balance() `function. **NOTE:** this means that unlike the current implementation, readlocks do not include accounts that already have a write lock

### Changes in emitted bytecode:
`@world` and `@acc allowing unbounded overdraft` sources now emit the `TAKE_ALWAYS` opcode instead of emitting the `TAKE_ALL` opcode. This allows them to send the required amount without checking if the balance is enough (which, in turn, allows the machine to avoid locking those accounts' balances)

### Changes in bytecode semantics
the `OP_TAKE_ALWAYS` now allows sending from an account even if it's balance is not found (this is the case with `@world` or with unbounded accounts)


components/ledger/internal/machine/script/compiler/compiler.go
```diff
@@ -26,6 +26,11 @@ type parseVisitor struct {
 	varIdx map[string]machine.Address
 	// needBalances store for each account, the set of assets needed
 	neededBalances map[machine.Address]map[machine.Address]struct{}
+
+	// like `sources`, but for accounts that aren't unbounded
```
gfyrag, line 30: aren't bounded no?

components/ledger/internal/engine/command/commander.go
```diff
@@ -126,22 +125,13 @@ func (commander *Commander) exec(ctx context.Context, parameters Parameters, scr
...
 
-		involvedAccounts, involvedSources, err := func() ([]string, []string, error) {
-			involvedAccounts, involvedSources, err := m.ResolveResources(ctx, commander.store)
-			if err != nil {
-				return nil, nil, NewErrCompilationFailed(err)
-			}
-
-			return involvedAccounts, involvedSources, nil
-		}()
+		readLockAccounts, writeLockAccounts, err := m.ResolveResources(ctx, commander.store)
```
ascandone, line 128: Note: this is the only usage of the `ResolveResources` method

components/ledger/internal/machine/script/compiler/compiler.go
```diff
@@ -674,12 +684,14 @@ func CompileFull(input string) CompileArtifacts {
...
-		neededBalances: make(map[machine.Address]map[machine.Address]struct{}),
-		sources:        map[machine.Address]struct{}{},
+		errListener:       errListener,
+		instructions:      make([]byte, 0),
+		resources:         make([]program.Resource, 0),
+		varIdx:            make(map[string]machine.Address),
+		neededBalances:    make(map[machine.Address]map[machine.Address]struct{}),
+		sources:           map[machine.Address]struct{}{},
+		writeLockAccounts: map[machine.Address]struct{}{},
+		readLockAccounts:  map[machine.Address]struct{}{},
```
ascandone, line 694: here the diff is weird because of golang's formatting, but the actual diff is this:
```diff
+ writeLockAccounts: map[machine.Address]struct{}{},
+ readLockAccounts:  map[machine.Address]struct{}{},
}
```

components/ledger/internal/machine/script/compiler/compiler.go
```diff
@@ -688,17 +700,24 @@ func CompileFull(input string) CompileArtifacts {
...
 	}
-	sort.Stable(sources)
+	sort.Stable(writeLockAccounts)
 
 	artifacts.Program = &program.Program{
-		Instructions:   visitor.instructions,
-		Resources:      visitor.resources,
-		NeededBalances: visitor.neededBalances,
-		Sources:        sources,
+		Instructions:      visitor.instructions,
```
ascandone, line 716: actual diff:
```diff
Instructions:   visitor.instructions,
Resources:      visitor.resources,
NeededBalances: visitor.neededBalances,
- Sources:        sources,
+ReadLockAccounts:  readLockAccounts,
+WriteLockAccounts: writeLockAccounts,
```

components/ledger/internal/machine/vm/machine.go
```diff
@@ -607,16 +611,19 @@ func (m *Machine) ResolveResources(ctx context.Context, store Store) ([]string,
 		m.Resources = append(m.Resources, val)
 	}
 
-	involvedAccounts := make([]string, 0)
-	involvedSources := make([]string, 0)
-	for _, accountAddress := range involvedAccountsMap {
```
ascandone, line 612: ⚠️ **IMPORTANT DIFF**
Instead of marking every possible account that appears within the program (directly or indirectly - e.g. the result of `meta()`) as readLock account, we only iterate through `m.Program.ReadLockAccounts`

components/ledger/internal/machine/vm/machine_test.go
```diff
@@ -528,6 +529,59 @@ func TestWorldSource(t *testing.T) {
 	test(t, tc)
 }
 
+func TestUnboundedSourceSimple(t *testing.T) {
```
ascandone, line 532: The purpose of these tests is to cover the case in which something is sent from an account whose balance was not given

gfyrag (approved):

PR #1694, opened 2024-09-16
feat(payments): isolate configs inside each connectors directory

paul-nicolas:
This PR adds config files per connector instead
of a global one. That way, every connector code is
confined within its directory

Fixes ENG-1387


laouji (approved):
Generally looks good although there are some check annotations at the bottom of the diff you might want to check out.

Any reason you picked json for the config language instead of for example yaml? Not an issue but yaml is slightly more human "maintainable".

PR #1695, opened 2024-09-16
feat(payments): introduce generic connector

paul-nicolas:
Fixes ENG-1368


components/operator/internal/resources/payments/deployments.go
```diff
@@ -120,8 +193,12 @@ func createFullDeployment(ctx core.Context, stack *v1beta1.Stack,
 							Args:          []string{"serve"},
 							Env:           env,
 							Image:         image,
-							LivenessProbe: applications.DefaultLiveness("http", applications.WithProbePath("/_health")),
+							LivenessProbe: applications.DefaultLiveness("http", appOpts),
 							Ports:         []v1.ContainerPort{applications.StandardHTTPPort()},
+							// TODO(polo): only for v3
+							SecurityContext: &v1.SecurityContext{
+								ReadOnlyRootFilesystem: pointer.For(false),
```
laouji, line 200: Why is this needed for v3?

components/operator/internal/resources/payments/deployments.go
```diff
@@ -298,3 +375,19 @@ func createGateway(ctx core.Context, stack *v1beta1.Stack, p *v1beta1.Payments)
...
+		return fmt.Errorf("invalid temporal uri: %s", temporalURI.String())
+	}
+
+	if temporalURI.Path == "" {
+		return fmt.Errorf("invalid temporal uri: %s", temporalURI.String())
+	}
+
+	if !strings.HasPrefix(temporalURI.Path, "/") {
+		return fmt.Errorf("invalid temporal uri: %s", temporalURI.String())
+	}
```
laouji, line 390: consider using the net/url package to normalise the url so that both with and without leading slash is accepted:

```suggestion
	u, err := url.JoinPath(temporalURI.Path, "/")
	if err != nil {
		return fmt.Errorf("invalid temporal uri: %s", temporalURI.String())
	}
	temporalURI = u
```

components/payments/internal/api/services/errors.go
```diff
@@ -38,3 +39,18 @@ func newStorageError(err error, msg string) error {
 		msg: msg,
 	}
 }
+
+func handleEngineErrors(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	switch {
```
laouji, line 48:
```suggestion
	switch {
	case err == nil:
			return nil
```

PR #1700, opened 2024-09-17
feat(all): Add scopes in OpenAPI files

flemzord:


Dav-14 (approved):

Earthfile
```diff
@@ -52,16 +52,9 @@ build-final-spec:
     SAVE ARTIFACT build/latest.json
 
 openapi:
-    FROM core+base-image
-    COPY . /src
```
gfyrag, line 56: please avoid that.
COPY . will copy the entire repository remotely.
And the SAVE ARTIFACT will copy the entire repository again.

PR #1707, opened 2024-09-19
feat(payments): introduce back again events

paul-nicolas:
Fixes ENG-1376


components/payments/internal/connectors/engine/engine.go
```diff
@@ -153,7 +158,51 @@ func (e *engine) UninstallConnector(ctx context.Context, connectorID models.Conn
...
-func (e *engine) CreateBankAccount(ctx context.Context, bankAccountID uuid.UUID, connectorID models.ConnectorID) (*models.BankAccount, error) {
+func (e *engine) ResetConnector(ctx context.Context, connectorID models.ConnectorID) error {
+	connector, err := e.storage.ConnectorsGet(ctx, connectorID)
+	if err != nil {
+		return err
+	}
+
+	// Detached the context to avoid being in a weird state if request is
+	// cancelled in the middle of the operation.
+	ctx, _ = contextutil.Detached(ctx)
```
laouji, line 169: If the work is detached from the parent context we'll also need to create an internal waitgroup that can advertise to main that engine events such as this one are finished so that it can execute a graceful shutdown (similar to how http.Server.Shutdown() is implemented).

(I suggest doing this is a separate ticket though)

components/payments/internal/connectors/engine/engine.go
```diff
@@ -229,6 +278,147 @@ func (e *engine) HandleWebhook(ctx context.Context, urlPath string, webhook mode
 	return nil
 }
 
+func (e *engine) CreatePool(ctx context.Context, pool models.Pool) error {
+	if err := e.storage.PoolsUpsert(ctx, pool); err != nil {
+		return err
+	}
+
+	ctx, _ = contextutil.Detached(ctx)
```
laouji, line 286: This is only creating one workflow. Does it need to be detached?

components/payments/internal/connectors/engine/engine.go
```diff
@@ -23,8 +23,13 @@ import (
 type Engine interface {
 	InstallConnector(ctx context.Context, provider string, rawConfig json.RawMessage) (models.ConnectorID, error)
 	UninstallConnector(ctx context.Context, connectorID models.ConnectorID) error
-	CreateBankAccount(ctx context.Context, bankAccountID uuid.UUID, connectorID models.ConnectorID) (*models.BankAccount, error)
+	ResetConnector(ctx context.Context, connectorID models.ConnectorID) error
+	ForwartBankAccount(ctx context.Context, bankAccountID uuid.UUID, connectorID models.ConnectorID) (*models.BankAccount, error)
```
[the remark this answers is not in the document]
paul-nicolas: Ah yes good catch !

components/payments/internal/connectors/engine/engine.go
paul-nicolas: Yes you're right, let me add a ticket for this one !

components/payments/internal/connectors/engine/engine.go
paul-nicolas: Since we already inserted the pool inside the database, I wanted to be sure that the event is sent also, hence the detached context. What do you think ?

laouji (approved):

PR #1708, opened 2024-09-20
fix(mangopay): fix webhooks handler

paul-nicolas:
Fixes ENG-1365

### DIFF
--- a/components/payments/internal/connectors/plugins/public/mangopay/webhooks.go
+++ b/components/payments/internal/connectors/plugins/public/mangopay/webhooks.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"net/url"
 	"os"
 	"time"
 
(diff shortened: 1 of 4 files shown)


components/payments/internal/connectors/plugins/public/mangopay/webhooks.go
```diff
@@ -116,12 +116,13 @@ func (p Plugin) createWebhooks(ctx context.Context, req models.CreateWebhooksReq
 
 	webhookURL := fmt.Sprintf("%s/api/payments/v3/connectors/webhooks/%s", stackPublicURL, req.ConnectorID)
 	for eventType, config := range webhookConfigs {
+		url := fmt.Sprintf("%s%s", webhookURL, config.urlPath)
```
laouji, line 119: maybe use `url.JoinPath` since it is designed for this purpose

paul-nicolas: Done thanks !

PR #1714, opened 2024-09-24
feat(payments): add storage tests and fix payments upsert

paul-nicolas:
Fixes ENG-1409

### DIFF
--- a/components/payments/internal/storage/accounts.go
+++ b/components/payments/internal/storage/accounts.go
@@ -161,7 +161,7 @@ func fromAccountModels(from models.Account) account {
 	return account{
 		ID:           from.ID,
 		ConnectorID:  from.ConnectorID,
-		CreatedAt:    from.CreatedAt,
+		CreatedAt:    from.CreatedAt.UTC(),
 		Reference:    from.Reference,
 		Type:         string(from.Type),
 		DefaultAsset: from.DefaultAsset,
@@ -176,7 +176,7 @@ func toAccountModels(from account) models.Account {
 		ID:           from.ID,
 		ConnectorID:  from.ConnectorID,
 		Reference:    from.Reference,
-		CreatedAt:    from.CreatedAt,
+		CreatedAt:    from.CreatedAt.UTC(),
 		Type:         models.AccountType(from.Type),
 		Name:         from.Name,
 		DefaultAsset: from.DefaultAsset,
--- a/components/payments/internal/storage/balances_test.go
+++ b/components/payments/internal/storage/balances_test.go
@@ -1,0 +1,480 @@
+package storage
+
+import (
+	"context"
+	"math/big"
+	"testing"
+
+	"github.com/formancehq/go-libs/bun/bunpaginate"
+	"github.com/formancehq/go-libs/logging"
+	"github.com/formancehq/go-libs/pointer"
+	"github.com/formancehq/go-libs/time"
+	"github.com/formancehq/payments/internal/models"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	defaultBalances = []models.Balance{
+		{
+			AccountID:     defaultAccounts[0].ID,
+			CreatedAt:     now.Add(-60 * time.Minute).UTC().Time,
+			LastUpdatedAt: now.Add(-60 * time.Minute).UTC().Time,
+			Asset:         "USD/2",
+			Balance:       big.NewInt(100),
+		},
+		{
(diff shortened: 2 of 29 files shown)


components/payments/internal/storage/bank_accounts.go
```diff
@@ -38,30 +38,48 @@ type bankAccount struct {
...
-		Exec(ctx)
-	if err != nil {
-		return e("insert related accounts", err)
+	if len(idsUpdated) > 0 {
+		_, err = tx.NewUpdate().
+			Model((*bankAccount)(nil)).
+			Set("account_number = pgp_sym_encrypt(?::TEXT, ?, ?)", toInsert.AccountNumber, s.configEncryptionKey, encryptionOptions).
+			Set("iban = pgp_sym_encrypt(?::TEXT, ?, ?)", toInsert.IBAN, s.configEncryptionKey, encryptionOptions).
+			Set("swift_bic_code = pgp_sym_encrypt(?::TEXT, ?, ?)", toInsert.SwiftBicCode, s.configEncryptionKey, encryptionOptions).
+			Where("id = ?", toInsert.ID).
```
gfyrag, line 67: On update, you should always update fields ONLY if it changed.
Since the function is an upsert, I guess account_number, iban, and swift_bic_code can already have the correct value.
So you should check for that in the where clause.
If you do not, the values will be rewritten even if it does not change. And it will involve all the write process of postgres, including reindexing.

paul-nicolas: Discussed IRL, we update only when the bank account is created ! We have some tests on that, so everything is good !

components/payments/internal/storage/accounts.go
```diff
@@ -161,7 +161,7 @@ func fromAccountModels(from models.Account) account {
 	return account{
 		ID:           from.ID,
 		ConnectorID:  from.ConnectorID,
-		CreatedAt:    from.CreatedAt,
+		CreatedAt:    from.CreatedAt.UTC(),
```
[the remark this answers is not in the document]
paul-nicolas: I created a ticket to change that everywhere, thanks for the advice !Will do it in another PR !

components/payments/internal/storage/balances_test.go
```diff
@@ -0,0 +1,480 @@
...
+	"github.com/formancehq/go-libs/logging"
+	"github.com/formancehq/go-libs/pointer"
+	"github.com/formancehq/go-libs/time"
+	"github.com/formancehq/payments/internal/models"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	defaultBalances = []models.Balance{
```
laouji, line 18: If you stored these in a map (`map[accID][]models.Balance`) you could reference them directly later in your tests rather than having to redefine `expectedBalances` in every test.

It might also make it more obvious why you expect certain datasets in your query balance requests.

components/payments/internal/storage/bank_accounts.go
```diff
@@ -38,30 +38,48 @@ type bankAccount struct {
...
 	// Insert or update the bank account
-	_, err = tx.NewInsert().
+	var idsUpdated []uuid.UUID
+	err = tx.NewInsert().
 		Model(&toInsert).
+		Column("id", "created_at", "name", "country", "metadata").
 		On("CONFLICT (id) DO NOTHING").
-		Exec(ctx)
+		Returning("id").
+		Scan(ctx, &idsUpdated)
```
laouji, line 56: Couldn't you use `sql.Result.RowsAffected()` and `sql.Result.LastInsertId()` from the Exec result instead of changing the query to include a return value?

components/payments/internal/storage/payments_test.go
```diff
@@ -0,0 +1,898 @@
...
+		p := models.Payment{
+			ID:            pID1,
+			ConnectorID:   defaultConnector.ID,
+			Reference:     "test1",
+			CreatedAt:     now.Add(-20 * time.Minute).UTC().Time,
+			Type:          models.PAYMENT_TYPE_TRANSFER,
+			InitialAmount: big.NewInt(100),
+			Amount:        big.NewInt(100),
+			Asset:         "USD/2",
+			Scheme:        models.PAYMENT_SCHEME_OTHER,
```
laouji, line 238: Is this test data realistic? I was under the impression that since the PSPpayment contains only the data from the adjustment, this base payment object will contain the same data (payment type, amount, etc) as the adjustment that is being appended.

laouji (approved):

PR #1722, opened 2024-09-26
feat(payments): add original reference in psp payment

paul-nicolas:


components/payments/internal/models/payments.go
```diff
@@ -10,6 +10,14 @@ import (
 
 // Internal struct used by the plugins
 type PSPPayment struct {
+	// Original PSP payment/transaction reference.
+	// In case of refunds, dispute etc... this reference should be the original
+	// payment reference. In case it's the first payment, it should be the same
+	// as the Reference.
+	// If we have no new reference for the payments when the status is refunded
+	// or disputed etc... we should use the same reference.
+	OriginalReference string
```
laouji, line 19: I might suggest something more like `ParentReference` because it makes it more explicit that there is a hierarchy. And then allow this to be blank if the transaction is the first in its series.

You could also add a convenience method like `HasParent() bool` if you dont' feel like putting `if payment.ParentReference == ""` all over the place

paul-nicolas: Done thanks !

components/payments/internal/connectors/plugins/public/currencycloud/payments.go
```diff
@@ -114,15 +114,15 @@ func transactionToPayment(transaction client.Transaction) (*models.PSPPayment, e
...
-		OriginalReference: transaction.ID,
-		Reference:         transaction.ID,
-		CreatedAt:         transaction.CreatedAt,
-		Type:              paymentType,
-		Amount:            amount,
-		Asset:             currency.FormatAsset(supportedCurrenciesWithDecimal, transaction.Currency),
-		Scheme:            models.PAYMENT_SCHEME_OTHER,
-		Status:            matchTransactionStatus(transaction.Status),
-		Raw:               raw,
+		ParentReference: transaction.ID,
```
laouji, line 117: Shouldn't it be empty now?

paul-nicolas: Done sorry about that

laouji (approved):

PR #1724, opened 2024-09-26
feat: (payments) add stripe v3 connector

laouji:
[ENG-1369](https://linear.app/formance/issue/ENG-1369/payments-v3x-add-stripe-connector)


components/payments/Earthfile
```diff
@@ -108,6 +108,15 @@ tidy:
     WORKDIR /src/components/payments
     DO --pass-args stack+GO_TIDY
 
+generate:
+    FROM core+builder-image
+    RUN apk update && apk add openjdk11
+    DO --pass-args core+GO_INSTALL --package=go.uber.org/mock/mockgen@latest
```
paul-nicolas, line 114: I have a question here:why changing the gomock package from github.com/golang/mock to go.uber.org/mock/mockgen?

components/payments/internal/connectors/plugins/public/stripe/client/client.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+
+	"github.com/stripe/stripe-go/v79"
+	"github.com/stripe/stripe-go/v79/account"
+	"github.com/stripe/stripe-go/v79/balance"
+	"github.com/stripe/stripe-go/v79/balancetransaction"
+	"github.com/stripe/stripe-go/v79/bankaccount"
+)
+
+//go:generate mockgen -source client.go -destination client_generated.go -package client . Client
```
paul-nicolas, line 13: Very nice, I'm gonna create a ticket to add it to every connectors !

components/payments/internal/connectors/plugins/public/stripe/accounts.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/formancehq/payments/internal/connectors/plugins/currency"
+	"github.com/formancehq/payments/internal/models"
+)
+
+var (
+	rootAccountReference = "root"
+)
+
+// root account reference is internal so we don't pass it to Stripe API clients
+func resolveAccount(ref *string) *string {
```
paul-nicolas, line 17: Do you need to pass a pointer here for the reference ?

components/payments/internal/connectors/plugins/public/stripe/accounts.go
```diff
@@ -0,0 +1,91 @@
...
+			Name:      &rootAccountReference,
+			Reference: rootAccountReference,
+			CreatedAt: time.Now().UTC(),
+			Raw:       json.RawMessage("{}"),
+			Metadata:  map[string]string{},
+		})
+		oldState.InitFinished = true
+	}
+
+	needed := req.PageSize - len(accounts)
```
paul-nicolas, line 50: The PageSize request parameters can be set by the user at connectors creation, and stripe "limit" query param is only valid if between 1 and 100. So if the users puts something > 100, all calls will return an error !

components/payments/internal/connectors/plugins/public/stripe/accounts.go
```diff
@@ -0,0 +1,91 @@
...
+
+	newState := AccountsState{
+		InitFinished: oldState.InitFinished,
```
paul-nicolas, line 53: Why not setting the LastID with the oldstate LastID ?

If the call to return accounts returns none (for example last page, but empty), you will reset the state and refetch everything since the LastID on the next call will be empty

components/payments/internal/connectors/plugins/public/stripe/client/payments.go
```diff
@@ -0,0 +1,51 @@
...
+		&expandSourceCharge,
+		&expandSourceDispute,
+		&expandSourcePayout,
+		&expandSourceRefund,
+		&expandSourceTransfer,
+		&expandSourcePaymentIntent,
+		&expandSourceRefundPaymentIntent,
+	}
+
+	itr := c.balanceTransactionClient.List(&stripe.BalanceTransactionListParams{
```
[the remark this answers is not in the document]
paul-nicolas: Actually, we're gonna have a problem, since the balance transactions are started from the most recent, you will fetch the whole history without problems, but when new payments are coming you will never fetch them since they are on the other side of the list

components/payments/internal/connectors/plugins/public/stripe/client/accounts.go
```diff
@@ -0,0 +1,26 @@
...
+func (c *client) GetAccounts(ctx context.Context, lastID *string, pageSize int64) ([]*stripe.Account, bool, error) {
+	// TODO
+	//	f := connectors.ClientMetrics(ctx, "stripe", "list_accounts")
+	//	now := time.Now()
+	//	defer f(ctx, now)
+	filters := stripe.ListParams{
+		Limit: &pageSize,
+	}
+	if lastID == nil {
+		filters.StartingAfter = lastID
```
paul-nicolas, line 18: Same reflexions as the balance transaction one

components/payments/internal/connectors/plugins/public/stripe/client/external_accounts.go
```diff
@@ -0,0 +1,39 @@
...
+) ([]*stripe.BankAccount, bool, error) {
+	// TODO
+	//	f := connectors.ClientMetrics(ctx, "stripe", "list_accounts")
+	//	now := time.Now()
+	//	defer f(ctx, now)
+	filters := stripe.ListParams{
+		Limit: &pageSize,
+	}
+	if lastID == nil {
+		filters.StartingAfter = lastID
```
paul-nicolas, line 23: Same reflexion as the payments comment

components/payments/internal/connectors/plugins/public/stripe/external_accounts.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+	}
+	if req.FromPayload == nil {
+		return models.FetchNextExternalAccountsResponse{}, errors.New("missing from payload when fetching external accounts")
+	}
+	if err := json.Unmarshal(req.FromPayload, &from); err != nil {
+		return models.FetchNextExternalAccountsResponse{}, err
+	}
+
+	newState := AccountsState{}
```
paul-nicolas, line 31: Same here, you should initialize with the old values in case

components/payments/Earthfile
paul-nicolas: My bad, as @gfyrag said, the former is deprecated !

components/payments/internal/connectors/plugins/public/stripe/client/payments.go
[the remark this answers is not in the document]
laouji: Ok I see now that the original implementation was using `created[lt]` to fetch balance transactions. Will see how I can replicate that with the SDK.

components/payments/internal/connectors/plugins/public/stripe/accounts.go
laouji: I don't need to. We can change it to a simple string if you prefer.

More just stylistic because the v2 implementation was passing around a potentially nillable account to methods like Translate that behave slightly different depending on whether an account was specified or not. So I am treating `from.Reference` from the cache as something that we expect might be nil across the board (as opposed to having empty string checks).

components/payments/internal/connectors/plugins/public/stripe/accounts.go
laouji: Oops. Good catch.

components/payments/internal/connectors/plugins/public/stripe/client/payments.go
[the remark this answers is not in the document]
laouji: I've updated it to fetch transactions, accounts and external accounts in ASC order

components/payments/internal/connectors/plugins/public/stripe/accounts.go
laouji: This is addressed

components/payments/internal/connectors/plugins/public/stripe/accounts.go
laouji: Created a ticke for this here: https://linear.app/formance/issue/ENG-1420/featpayments-validate-page-size-in-connectors

paul-nicolas (approved):

PR #1727, opened 2024-09-27
fix(payments): fix docker compose and banking circle connector

paul-nicolas:
Fixes ENG-1364


components/payments/internal/connectors/plugins/public/bankingcircle/client/auth.go
```diff
@@ -46,7 +46,6 @@ func (c *Client) login(ctx context.Context) error {
 		}
 		return fmt.Errorf("failed to login: %d", statusCode)
 	}
-	return fmt.Errorf("failed make login request: %w", err)
```
laouji, line 49: 🙈

laouji (approved):

PR #1734, opened 2024-10-18
fix(balances): fix balances insertion

paul-nicolas:


components/payments/cmd/connectors/internal/storage/balances.go
```diff
@@ -2,59 +2,119 @@ package storage
...
-		(SELECT *
-			FROM cte1
-			WHERE cte1.balance != COALESCE((SELECT balance FROM accounts.balances WHERE account_id = cte1.account_id AND last_updated_at < cte1.last_updated_at AND currency = cte1.currency ORDER BY last_updated_at DESC LIMIT 1), cte1.balance+1)
-		) data`)
+	tx, err := s.db.BeginTx(ctx, &sql.TxOptions{})
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = tx.Rollback()
```
laouji, line 22: Rolling back outside of a transaction (eg. in the case that `tx.Commit()` was successful) results in `noop` and not an error, so you shouldn't have to discard this error. Let's log it perhaps?

components/payments/cmd/connectors/internal/storage/balances.go
```diff
@@ -2,59 +2,119 @@ package storage
...
+	return e("failed to commit transaction", tx.Commit())
+}
 
-	_, err := query.Exec(ctx)
+func (s *Storage) insertBalances(ctx context.Context, tx bun.Tx, balance *models.Balance, checkIfAccountExists bool) error {
+	var account models.Account
+	err := tx.NewSelect().
+		Model(&account).
+		Where("id = ?", balance.AccountID).
+		Scan(ctx)
```
laouji, line 39: Shouldn't we skip running this query at all if `checkIfAccountExists` is false? Or else, what is the purpose of this flag?

Previously it was only checking if the account existed in `accounts.account` if this flag is true.

paul-nicolas: The purpose of this flag is to not insert the balance if the account is not existing, so we need this check before I think

components/payments/cmd/connectors/internal/storage/balances.go
paul-nicolas: Done thanks

components/payments/cmd/connectors/internal/storage/balances.go
laouji: It doesn't look like that to me.

The crucial difference between the two SQL statements is whether `WHERE EXISTS (SELECT 1 FROM accounts.account WHERE id = cte1.account_id)` is run or not.
So if `checkIfAccountExists` is true, and the above SQL doesn't return results, then the balance insertion will be skipped.

However if `checkIfAccountExists` is false it doesn't care if there is a corresponding entry in `accounts.account` or not, and will proceed with the balance update as long as it can find entries that match `balance.account_id` that meet the criteria.

I have to admit I don't fully understand the use-case though. In what cases do we have `accounts.balance.account_id` that don't correspond to records in `accounts.account` ?

laouji: Since there is a foreign key on the db, I understand now what you mean about how there would have been an error in the original implementation. That resolves my comment.

laouji (approved):

PR #1759, opened 2025-04-01
fix(events): fix payments v3.0.0 events names

paul-nicolas:


Dav-14 (approved):

laouji (approved):

paul-nicolas:
Canceled this one as we will use the version and add some docs about upgrading to v3

PR #1763, opened 2025-04-03
fix(payments): fix payment storage constraint

paul-nicolas:
We want to have an unique contraint on the reference of payments within the same connector, but not accross connectors

### DIFF
--- a/components/payments/internal/storage/migrations_v1.x.go
+++ b/components/payments/internal/storage/migrations_v1.x.go
@@ -396,6 +396,21 @@ func registerMigrationsV1(ctx context.Context, migrator *migrations.Migrator) {
 				return fixMissingReferenceTransferInitiation(ctx, tx)
 			},
 		},
+		migrations.Migration{
+			Up: func(tx bun.Tx) error {
+				// Rererence should be unique within a connector, but not accross
+				// connectors. This migration fixes it.
+				_, err := tx.Exec(`
+					ALTER TABLE payments.payment ADD CONSTRAINT payment_reference_connector_id_unique_key UNIQUE (reference, connector_id);
+					ALTER TABLE payments.payment DROP CONSTRAINT IF EXISTS payment_reference_key;
+				`)
+				if err != nil {
+					return err
+				}
+
+				return nil
+			},
+		},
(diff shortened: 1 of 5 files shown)


components/payments/internal/storage/migrations_v1.x.go
```diff
@@ -396,6 +396,21 @@ func registerMigrationsV1(ctx context.Context, migrator *migrations.Migrator) {
 				return fixMissingReferenceTransferInitiation(ctx, tx)
 			},
 		},
+		migrations.Migration{
+			Up: func(tx bun.Tx) error {
+				// Rererence should be unique within a connector, but not accross
+				// connectors. This migration fixes it.
+				_, err := tx.Exec(`
+					ALTER TABLE payments.payment ADD CONSTRAINT payment_reference_connector_id_unique_key UNIQUE (reference, connector_id);
+					ALTER TABLE payments.payment DROP CONSTRAINT IF EXISTS payment_reference_key;
```
laouji, line 405: How much data does the client have in this table? I think this is likely put an access exclusive lock on the table which could cause some downtime.

If you put the two statements in separate Exec blocks you might be able to to use the `CONCURRENTLY` keyword to avoid locking the full table.

laouji: Seems like the preferred strategy is to add a unique index on the field with the `CONCURRENTLY` keyword then run an ADD constraint using that new index. Since the index already guarantees the uniqueness of the constraint, the whole table is not locked when the constraint is added.

From the docs:

> Adding a constraint using an existing index can be helpful in situations where a new constraint needs to be added without blocking table updates for a long time. To do that, create the index using CREATE UNIQUE INDEX CONCURRENTLY, and then convert it to a constraint using this syntax.

source:
https://www.postgresql.org/docs/current/sql-altertable.html

paul-nicolas: @laouji payments v2 is still on the old migration system and we have transaction in arg, so we can't do that unfortunately without migrating the whole payments v2 to the new migration system

laouji (approved):

PR #1979, opened 2026-02-26
feat: add transaction plane service

paul-nicolas:


flemzord:
Closed automatically: bulk stale PR cleanup

